PR #16, opened 2024-01-27
Updated all mikro-orm packages to same version

JordanPlayz158:
Updated mikro-orm packages from version 5.7.12 to 5.9.7 to fix an issue from only one of the packages being on 5.9.7, all need to be on same version or mikro-orm will warn you and exit causing the server to be unable to start.


JordanPlayz158:
Don't merge yet, GHark wishes to bump to v6 of Mikro

React1-X:
Can this be closed? Or are we still looking into bumping versions for packages? @JordanPlayz158

JordanPlayz158:
Well we need to at least make sure that all mikro-orm packages are the same version otherwise the server does not start without changing it manually. As we are looking to go to v6, there may be breaking changes that need to be looked into, will try out the v6 change when I have the time as it should be easy to test.

BigmanGHark:
> Can this be closed? Or are we still looking into bumping versions for packages? @JordanPlayz158

We can merge this in the meantime if you would like- I took a look at v6 after and I saw that some of the updates might be a pain, its a long list https://mikro-orm.io/docs/upgrading-v5-to-v6

JordanPlayz158:
Ok, for v6, going one issue at a time, we have a decision though, do we wish to keep our current ENV names (like DB_NAME) and manually load the `.env` due to a v6 breaking change https://github.com/mikro-orm/mikro-orm/discussions/5178 or do we wish to just change the `.env` variable names to the MIKRO_ORM_DB_NAME and such?

JordanPlayz158:
That was not my intended goal, when we test this and confirm it is ready, I'll probably remake the branch so the git history doesn't get mangled

React1-X:
Hey @JordanPlayz158 - I think I'll close this PR for now, as it's quite old / out-of-sync. and everything seems to be fine in terms of not needing any drastic upgrades yet (e.g. we aren't using anything too critical with the ORM)

However, when we decide to upgrade at a later stage, we can re-create this PR with all the newest changes. Thanks,

PR #18, opened 2024-01-29
feature: asset cdn

sush1lemon:
- use different url for assets

### DIFF
--- a/client/scripts/com/monsters/display/ImageCache.as
+++ b/client/scripts/com/monsters/display/ImageCache.as
@@ -7,6 +7,7 @@ package com.monsters.display
    import flash.events.IOErrorEvent;
    import flash.events.TimerEvent;
    import flash.net.URLRequest;
+   import flash.system.LoaderContext;
    import flash.utils.Timer;
    
    public class ImageCache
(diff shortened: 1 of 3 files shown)


BigmanGHark:
We probably need to handle this in our crossdomain in future (although maybe its lax enough it wont matter), other than that looks good

BigmanGHark (approved):

client/scripts/com/monsters/display/ImageCache.as
```diff
@@ -208,7 +209,7 @@ package com.monsters.display
          var l:Loadable = queue;
          l.loadState = LOADING;
          req_str = l.shouldPrepend ? prependImagePath + l.key : l.key;
-         l.loader.load(new URLRequest(req_str));
+         l.loader.load(new URLRequest(req_str), new LoaderContext(true));
```
BigmanGHark, line 212: Out of curiosity whats this do?

sush1lemon: it's for the crossdomain.xml, just telling the load function to the get crossdomain.xml of the cdn url. Like in the [AuthForm](https://github.com/bym-refitted/backyard-monsters-refitted/blob/9c0227998db95b4bbba5e517e4bd0c9947e73d5f/client/scripts/AuthForm.as#L138)

BigmanGHark: Nice one thank you

BigmanGHark:
@React1-X Could we merge this one. This seems like all the client side changes we need for now and we can update the server url later

React1-X:
> @React1-X Could we merge this one. This seems like all the client side changes we need for now and we can update the server url later

Yeah sure, sounds good to me. Merging.

PR #23, opened 2024-02-01
Added log switches for local env

React1-X:


BigmanGHark (approved):

BigmanGHark:
LGTM, a question about logging user ID but we can merge and do that another time

PR #29, opened 2024-02-06
[POC] Use socket for base save/update | Prevent multiple session

sush1lemon:
- Add tcp server to prevent same account multiple flash intance
- Uses redis to cache the user id with expiration (15s)


BigmanGHark:
Will this work in the browser if we ever support that? Just curious about how that would work - think we may need to use Websockets instead of a raw tcp connection for browser unless we just dont care about ever supporting that cc: @React1-X 

Maybe there is a way to do a tcp connection via webRTC protocol but IDK

sush1lemon:
I think ruffle provide a workaround for this: [ruffle socket](https://github.com/ruffle-rs/ruffle/wiki/Frequently-Asked-Questions-For-Users#how-can-i-connect-to-a-tcpsocket-or-xmlsocket-from-the-web). Or we can just implement a websocket version once we start using browser

tram98:
@sush1lemon @BigmanGHark I'd suggest to use a WebSocket instead. It also would make sense to refactor the URLLoaderAPI to work more service oriented.

React1-X:
> @sush1lemon @BigmanGHark I'd suggest to use a WebSocket instead. It also would make sense to refactor the URLLoaderAPI to work more service oriented.

I think this has been discussed before, but, the easiest solution is:

- Store a user's session / token in the database, I believe we are doing 90% of this already except for the actual storing - and then when they try to login again on another Flash window, the token will be invalid because it will be different to the one stored in the database. This avoids the complexities of using AS3 WebSockets, and probably reduces overhead.

React1-X:
This issue has been solved using simple token invalidation on the server - however, I will keep this PR here as POC, just in case.

PR #41, opened 2024-03-18
Linux Launcher + Update Instructions

hideki2k02:
Took me a few days to realize i've yet to submit this PR, sorry ReactX lol

I've also included the link to the Archive page with the Mac/Linux Binaries and I've renamed the previous "How to play" Instructions as Manual Installation

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,10 +20,18 @@ This repository contains the entire modified SWF (Shockwave Flash) client files,
 <br>
 
 ## 🚀 How to Play
-We recommend to run the game in the native Flash Player projector, which is included in the root of this repository [here](./flashplayer_32.exe). Once you have downloaded and opened the Flash Player, click on `File > Open` and select the game file. You can find the game file under our [latest releases](https://github.com/bym-refitted/backyard-monsters-refitted/releases).
+
+### Linux
+You can use [this](./launcher.sh) script to automatically set-up the game for you: It checks for the latest release and automatically downloads the Flash Player if not found.
+
+### Manual Installation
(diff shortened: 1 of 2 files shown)


EvenBiggerManGhark:
I was thinking ideally we would accept some user input to choose a version of the game to launch but that has the potential to get annoying, would like to hear your thoughts on that?

hideki2k02:
> I was thinking ideally we would accept some user input to choose a version of the game to launch but that has the potential to get annoying, would like to hear your thoughts on that?

I think the end-user will  just be annoyed. Only Devs and a very specific group of people would prefer a Nightly Builds that could potentially break the save. 

However, I do think something like that could be included on the launch script in the future. Here's an example: 
(Written in toml for easier reading)
```toml
[general]
check_for_updates=true
update_channel='bymr-stable'

[dev]
# Overrides check_for_updates
forced_version='0.2.4'
```

EvenBiggerManGhark:
Think there are a good few local users with people running their own servers, but now that I think about it, they should probably manage that themselves given they will have to update their backend also. The only thing that people might need at the moment is the http version as a very small amount of users can not access over http.

TOML would be ideal but the thought of trying to make batch / powershell work with that makes me feel physically ill right now. We can chalk that down as an idea for another day but good to have in mind. Have the windows one launching, once its fully finished up I might adapt yours for Mac users as well.

But yeah I agree that for now we can just stick with launching stable and worry about that later 👍

EvenBiggerManGhark:
We can move this to the launcher repo - have added the windows script there. Think we can still keep these as alternatives

hideki2k02:
Will do! I can see that a certain user-base will prefer the Script version, especially tinkerers!

PR #46, opened 2024-03-22
[WIP] Inferno Integration

DedProtectr:


server/src/mikro-orm.config.ts
```diff
@@ -4,21 +4,22 @@ import path from 'path';
...
 
 // The configuration for the ORM - Any Entities added need to be put in here, other than that probably doesn't need to be touched
 // tslint:disable-next-line: no-object-literal-type-assertion
 export default {
-  type: 'mariadb',
+  type: "mariadb",
   allowGlobalContext: false,
-  debug: true,
-  entities: [Save, User, WorldMapCell],
+  debug: false,
```
React1-X, line 14: Can we please revert this change for this PR?

server/package.json
```diff
@@ -28,6 +28,7 @@
     "bcrypt": "^5.1.1",
     "chalk": "^5.3.0",
     "dotenv": "^16.3.1",
+    "install": "^0.13.0",
```
React1-X, line 31: Can we remove this? It doesn't seem to be used anywhere across the codebase

React1-X: We should also revert the changes to `package-lock.json`

server/src/controllers/baseLoad.ts
```diff
@@ -33,12 +34,18 @@ export const baseLoad: KoaController = async (ctx) => {
...
       throw saveFailureErr;
     }
+  } else if (requestBody.type === "ibuild") {
+    save = await loadBuildBase(ctx, authSave.baseid_inferno.toString());
+    if (save && save.saveuserid !== user.userid) {
+      throw saveFailureErr;
+    }
   } else {
     save = await loadViewBase(ctx, requestBody.baseid)
   }
```
EvenBiggerManGhark, line 51: We could clean this up, not sure why we had this in the first place:
```javascript
let baseId = requestBody.baseid;
if (requestBody.type === "ibuild") {
baseId = authSave.baseid_inferno.toString()
}

save = await loadBuildBase(ctx, baseId);
if (save && save.saveuserid !== user.userid) {
  throw saveFailureErr;
}
```

If there is some reason we did not do error handling on non build requests we could just modify the last part to 
```javascript
if (requestBody.type.includes("build") && save && save.saveuserid !== user.userid) {
  throw saveFailureErr;
}
```

EvenBiggerManGhark: Sorry let me fix comment above misread something

PR #47, opened 2024-03-27
Add Check to Make Sure Maproom Finished Initializing (Crash Fix)

TheClassic0437:
Fixes a bug where opening the maproom just after placing it for the first time would crash the game.

### DIFF
--- a/client/scripts/MAPROOM.as
+++ b/client/scripts/MAPROOM.as
@@ -130,7 +130,7 @@ package
                   }
                   else if(!_initFinishedMR)
                   {
-                     GLOBAL.Message(KEYS.Get("The Maproom is still initializing. Please wait a moment before opening it."));
+                     GLOBAL.Message(KEYS.Get("newmap_init_setup"));
                   }
                   else
                   {
--- a/client/scripts/MAPROOM.as
+++ b/client/scripts/MAPROOM.as
@@ -1,6 +1,5 @@
 package
 {
-   
    import com.monsters.ai.TRIBES;
    import com.monsters.ai.WMBASE;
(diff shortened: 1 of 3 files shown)


client/scripts/MAPROOM.as
```diff
@@ -126,6 +128,10 @@ package
                         ShowB();
                      }
                   }
+                  else if(!_initFinishedMR)
+                  {
+                     GLOBAL.Message(KEYS.Get("The Maproom is still initializing. Please wait a moment before opening it."));
```
React1-X, line 133: Can we change this hardcoded text to come from the language file located in `server/public/gamestage/assets/en.v8.json`.

Let's create a new key with this name and this text on line 1648, just before `newmap_bm_name`:

```json
"newmap_init_setup": "Please wait a moment while we set up the MapRoom. This should only take a few seconds"
```

So now you can call `newmap_init_setup` in your message, rather than the hardcoded text.

React1-X:
Everything else works great! Thanks for this change. @TheClassic0437

client/scripts/MAPROOM.as
TheClassic0437: > Can we change this hardcoded text to come from the language file located in `server/public/gamestage/assets/en.v8.json`.

Yep!
I've made the change you requested. It should be working now!
I left it hardcoded initially from not knowing where the language files were, apologies.

React1-X: Great, thanks! Just one last thing, can we add a comma to the end of that JSON that you added in `en.v8.json`, otherwise it's not valid JSON. ^^

React1-X (approved):

PR #77, opened 2024-10-29
Minor fixes

tram98:
**YARD_PROPS**: Resolved duplicate keys by either removing them or changing them to their intended values


tram98:
Hey @EvenBiggerManGhark ,

That's not a dumb question.

The MovieClip patch is only CleanUp.

The YARD_PROPS is an actual bugfix (including cleanup of duplicate keys), as it fixes the reference to some images.

tram98:
@EvenBiggerManGhark I saw lots of function stubs in `PATHING` referring to rendering, so I assume that the class was at some point during debugging added to the stage to display debug information. Unfortunately, there is no code present, so we do not know what debugging was used.

React1-X (approved):

PR #84, opened 2024-11-08
Fix: Store item expiry calculation

iilegacyyii:
Store item expiry is expected to be epoch (seconds). It was being calculated as `currentTime (milliseconds) + duration (seconds)`.

This has been corrected.


React1-X:
Good spot! Approved + merging

PR #85, opened 2024-11-11
Fully rewritten the GetCellsInRange function

Mateo-os:
This function determines which cells of the map room the player can attack. As you can see, Kixeye literally inputted those values by hand. This PR changes it to find the cells algorithmically in a more easy-to-debug way.
Also, the variable names of the GetCell function were renamed so they would not be decompiler-generated names.

Note: I didn't check if Outposts's range expansion works, but considering the function only takes the center coordinates and a range in units, it should work.


React1-X:
## Discord Conversation on this PR:

### **React:** 
Thanks a lot for this @Mateo-os 😄

When we get a chance, myself and Ghark will take a look.

Quick question: how thoroughly have you tested it? And is it a 1:1 of the original functionality? (just not using Kixeye's shitty implementation)

<br>

### **Matnamalo:** 
It should be the original functionality 
I made some (very, very barebones) simulations in python to test it and it looked fine.
What i didn't test and just now i'm realizing is how it works with expanding range via outposts. That being said the function just takes one pair of x , y coordinates and a range, so it probably is being called once for each outpost, and shouldn't cause issues. 

<br>

### **React:** 
Okay great! I will take a look deeper when I find time, I guess the one thing i can spot off the bat is that it's written very "Pythony" xD

Probably just me being picky, but maybe some formatting / camelCase variables and just more general JS-styled namings and stuff 😄

Other than that, looks good from the eye-test

<br>

### **Matnamalo:** 
haha you caught me

<br>

### **React:** 
Haahaha it's okay, let's just try to keep the AS3 standard, also, are we forced into 3 nested for loops? Is this how they did it? Can we optimize this?

<br>

### **Matnamalo:**
They didn't use any loops, they setted the variables by hand
Like the first 120 maproom cells are hardcoded
And then that code is DUPLICATED because how the map works, the neighbours of cells with even X must be calculated different that the neighbours of cells with odd X
The outer loop is so the cells are added in order from closer to farther away. 
I COULD remove it, but 
A) Don't Know how it would impact the code (When i implemented i was dealing with another segfault issues istg this code) 
B) Kixeye's version is organized closer to farthest

<br>

### **React:**
I think we can find a way to optimize the logic, generally 3 for-loops is quite some overhead, or at least we can break it down and make it a little more suitable, because I guess the performance is critical on the MapRoom (especially cell calculation) so we need to be very careful on this one. 

But yeah, for now I would suggest the AS3 / JS styled approach, maybe some additional comments on what the code is doing rather than the system being used, other than that it is fine, thank you! 😄

Mateo-os:
@EvenBiggerManGhark, performance-wise, the original code was already very good, although very messy to read as it hardcoded the first 120 cells. That being said, for what I tested, I noted no performance loss with the rewrite.

Mateo-os:
I closed the PR because i messed up when squashing the commits, I'll open a new one with a clean branch.

PR #88, opened 2024-11-17
fixed a syntax error in try-init-db.sh

tram98:


tram98:
I have no idea how this has not failed earlier on my machine, but I found the error just now, so here is the fix.

React1-X (approved):

PR #93, opened 2024-11-19
bugfix: multiple bug fixes

sush1lemon:
- fix outpost cycling resource bug
- fix upgrade bug when viewing outpost
- fix outpost monster not saving when used in an attack


server/src/services/base/updateMonsters.ts
```diff
@@ -16,7 +16,7 @@ export const updateMonsters = async (monsterupdates: MonsterUpdate[]) => {
   // Iterate over the saves and apply the updates
   for (const save of saves) {
     const monsterUpdate = monsterupdates.find(
-      (update) => update.baseid === save.baseid
+      (update) => update.baseid == save.baseid
```
sush1lemon, line 19: **save.id** is a string while the **update.baseid** is a number, idk why

React1-X (approved):

PR #97, opened 2024-11-26
Resolved townhall resolution mechanism in balancedReward.ts.

tram98:
@React-1 just citing your message...

> Description:
>
>This fix will help reduce a lot of tickets, it has to do with user's not getting rewards at certain town hall levels, such as Korath, Krallen, >Diamond Spurtz Cannon.
>
>Details:
>
>These rewards are handled on the server, in the file balancedRewards.ts
>
>The issue comes from the fact that the script checks for the first key in the object being "0" since this building should always be the Town >Hall, and checks the "l" property on that object being the level of the Town Hall. If its above a certain level, insert the reward into the >database.
>
>The oversight was trusting the Backyard Monsters game client to always make sure the first building is the Town Hall. For some odd reason, the >people who report the issue, don't have a first key ("0") in their buildingdata, instead their Town Hall is some random key like "1776"

Also see my comment below regarding the champion data code changes

### DIFF
--- a/server/src/controllers/base/load/balancedReward.ts
+++ b/server/src/controllers/base/load/balancedReward.ts
@@ -54,13 +54,15 @@ export const balancedReward = async (userSave: Save) => {
   }
 
   if (townHall && townHall.l >= 7) {
-    let champ = userSave.champion;
+    // sometimes the champion variable is an object instead of a string,
+    // so the JSON.parse call runs into an error that "[object Object]" is not valid JSON.
+    let championRawData = userSave.champion;
     let championData = [];
-    if(typeof champ === "string") {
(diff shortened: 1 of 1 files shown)


tram98:
@React-1 regarding the champion object related changes: I had this error happen to me on my local system, so I added that check. 

Our server seems to be very sensitive with the environment it is running in. I had to mode away from hosting the server inside docker because of errors that only happen INSIDE the container.

tram98:
I merged the wrong branch. DO NOT MERGE THIS PR

tram98:
This PR is replaced by https://github.com/bym-refitted/backyard-monsters-refitted/pull/98

PR #98, opened 2024-11-26
Bugfix - balancedReward 

tram98:
@React-1 just citing your message...

> Description:
>
>This fix will help reduce a lot of tickets, it has to do with user's not getting rewards at certain town hall levels, such as Korath, Krallen, >Diamond Spurtz Cannon.
>
>Details:
>
>These rewards are handled on the server, in the file balancedRewards.ts
>
>The issue comes from the fact that the script checks for the first key in the object being "0" since this building should always be the Town >Hall, and checks the "l" property on that object being the level of the Town Hall. If its above a certain level, insert the reward into the >database.
>
>The oversight was trusting the Backyard Monsters game client to always make sure the first building is the Town Hall. For some odd reason, the >people who report the issue, don't have a first key ("0") in their buildingdata, instead their Town Hall is some random key like "1776"

Regarding the champion object related changes: I had this an error happen to me on my local system, so I added a check that addresses it. See the comment for more details.

In general, our server seems to be very sensitive to the environment it is running in. I had to mode away from hosting the server inside docker because of errors that only happen INSIDE the container.

### DIFF
--- a/server/src/controllers/base/load/balancedReward.ts
+++ b/server/src/controllers/base/load/balancedReward.ts
@@ -54,13 +54,15 @@ export const balancedReward = async (userSave: Save) => {
   }
 
   if (townHall && townHall.l >= 7) {
-    let champ = userSave.champion;
+    // sometimes the champion variable is an object instead of a string,
+    // so the JSON.parse call runs into an error that "[object Object]" is not valid JSON.
+    let championRawData = userSave.champion;
     let championData = [];
-    if(typeof champ === "string") {
+    if(typeof championRawData === "string") {
       championData = JSON.parse(userSave.champion || "[]");
     }
     else {
-      championData = champ;
+      championData = championRawData;
     }
 
     if (Object.keys(userSave.krallen).length === 0) {
--- a/server/src/controllers/base/load/balancedReward.ts
+++ b/server/src/controllers/base/load/balancedReward.ts
@@ -15,8 +15,6 @@ import { ORMContext } from "../../../server";
  * @throws Error when townHall Object cannot be found
  */
 function parseTownhallFromBuildingData(buildingData: FieldData) {
-  let townHall: any = null;
-
   for (const key in buildingData) {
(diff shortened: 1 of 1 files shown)


server/src/controllers/base/load/balancedReward.ts
```diff
@@ -1,7 +1,36 @@
...
+  let townHall: any = null;
+  for(const key in buildingData) {
+    const building = buildingData[key];
+    if(!building) {
+      continue;
+    }
+    if (building.t !== 14) continue;
+
+    townHall = building;
+    break;
```
senninr, line 27: line 21 - 27 can be simplified to
```
if (building && building.t === 14) {
  return building;
}
```
and by doing that line 18 & line 29 can be removed

senninr (changes requested):

server/src/controllers/base/load/balancedReward.ts
tram98: Indeed much cleaner! Thanks!

tram98 (commented):
I have implemented your simplification suggestion

server/src/controllers/base/load/balancedReward.ts
```diff
@@ -1,7 +1,46 @@
...
+ * 
+ * The object is identified by checking the `t` (type) variable on the child objects. 
+ * If `t` is equal to 14, then it is considered a town hall.
+ * 
+ * @param buildingData the buildingData object to search
+ * @returns the townHall object found
+ * @throws Error when townHall Object cannot be found
+ */
+function parseTownhallFromBuildingData(buildingData: FieldData) {
+  let townHall: any = null;
```
senninr, line 18: this unused variable can be removed

tram98: done

tram98:
Done @React1-X

React1-X (approved):

PR #103, opened 2024-12-03
SecNum implementation for build/upgrade times

AMXBus:


client/scripts/BUILDINGOPTIONSPOPUP.as
```diff
@@ -198,7 +198,8 @@ package
                   "r2":0,
                   "r3":0,
                   "r4":0,
-                  "r5":0
+                  "r5":0,
+                  "time":new SecNum(_loc3_.time.Get())
```
React1-X, line 202: @AMXBus - can you explain why we add a new time property onto this? My question around this is, if the original game didn't get the time from here, then where did it get from? And why do we give it a time?

client/scripts/BFOUNDATION.as
```diff
@@ -3170,7 +3170,7 @@ package
             {
                _loc1_.r4over = true;
             }
-            _loc1_.time = _loc2_.time;
+            _loc1_.time.Set(_loc2_.time.Get());
```
React1-X, line 3173: Just a question for my own understanding - why do we remove the assignment here and directly call `.Set()` on the time property?

Could this work:

```javascript
_loc1_.time = new SecNum(_loc2_.time);
````

File: client/scripts/GLOBAL.as [hunk 3]
```diff
@@ -572,114 +572,114 @@ package
...
-                  "r1": 1247840,
-                  "r2": 1247840,
-                  "r3": 623920,
-                  "r4": 0,
-                  "time": 97200,
-                  "re": [[14, 1, 4], [11, 1, 1]]
-               }];
-         _buildingProps[4].hp = [4000, 8000, 16000, 28000, 28000];
-         _buildingProps[4].capacity = [500, 1000, 1750, 2250, 3000, 4000, 4000];
+         _buildingProps[8].costs = [{
```
React1-X, line 575: Looks good! Let's just be careful with formatting changes.

client/scripts/BUILDINGOPTIONSPOPUP.as
AMXBus: During investigation on where all the time properties get called there are a few areas where if the time property was not ever added or got overwritten it would have a resolution on it's own. When I implemented SecNum however in all those calls many submenus would just refuse to work. Even some that get filled at runtime would stop the game with an exception.

This was most likely me searching over all mentions and usages of the time property of buildings get used and covering all bases.

client/scripts/GLOBAL.as
AMXBus: Sorry for the formatting mess. I was flipflopping between VSCode and JPEXS and never sat down to clean it up.

React1-X (approved):

PR #110, opened 2024-12-10
Implement SecNum for monster stats damage/HP

AMXBus:
Another SecNum change but these are a bit more complex.

Main things to note:

- We implement SecNum calls in the runtime intit structure just like before for the monster damage/HP stats (currently in CREATURELOCKER.as and RebalancedCreatures.as if we ever want to use those)
- The calls are a little bit different here as we incorporate the champion monsters as well (these stats are in CHAMPIONCAGE.as)
- This is the first moment where I have seen pointers (primitive but it's something) being used in this game as CHAMPIONCAGE.as and CREATURELOCKER.as have their own getter functions for properties. (CHAMPIONCAGE.GetGuardianProperty() and CREATURES.GetProperty())
- All POPUP equivalents use these property getters that are essentially full object passers so we need to be careful if we plan any enhancements further down the line.
- Some buildings use these monster properties for their own functions like bunkers and champion cage healing (housing as well if MR3 is active)
- There is a lot of complex math that gets involved at every tick() but is well secured with prior SecNum implementation
- This should not interfere with how the game calculates values for buffs/armor/dmg reduction as all that math is well in the background and works on each tick(). If any changes would happen mid calculation it would either crash the game or correct itself on the next one as these values get recalculated... all the time...


client/scripts/CHAMPIONCAGE.as
```diff
@@ -480,6 +480,11 @@ package
          {
             var _loc2_:Object = _guardians["G" + param1];
             var _loc3_:Object = _loc2_.props;
+            for (var temp = 0; temp < _loc3_.health.length; temp++)
```
React1-X, line 483: Hey @AMXBus - would you mind explaining the reason for introducing this loop here? I noticed it's also called "temp", is this accidentally included in the PR?

PR #123, opened 2025-01-17
Fix timer when fortifying buildings at max level

AMXBus:
A OG developer oversight that might have been in the game till it's very end. It would seem that they did not account for anything related to fortification when it came to the resource animation. When a building would get fortified it would grab the next level info for upgrades and the animation. This fails when buildings would be max level since there is no data for the next level. 

In an effort to reduce complexity of a simple cosmetic operation this fix has been implemented so that when the NaN (not a number)/null upgrade data get's detected it fills it with the proper fortify data. 

Functionally less work to be done and less areas of future fault, cosmetically fortify resource animation will contain the right amounts and types of resources only when the buildings are at max level (If people care or we are so far ahead where we need to worry about fixing the cosmetic parts of the game we'll change it then).


React1-X (approved):
LGTM!

PR #126, opened 2025-02-03
[QoL Enhancement] Resizeable maproom with support up to HD resolution

AMXBus:
Resize logic and function were obviously there but never implemented. 

- Added the resize function and handler to the Update() event function. 
- Setup resize logic and smoothing so that the UI elements do not jitter all on their own when resizing (currently happens with all UI elements that are independent of each other or their parent).
- Expanded cell container cells on both axis to support up to HD resolutions.

Currently buggy:

- View blockers when popups show up (they are the same resolution as the screen when the popups happen and gray out the screen but do not resize so it is possible to resize the screen and click on something else).
- Small dynamic UI elements like bookmarks do not update coordinates properly as they are not initialized as objects at the start of the map.
- Ofc huge frames start dropping when we want to fast scroll on larger resolutions.


React1-X:
> @React1-X I didn't work off of the full screen logic for that and it did seem broken at the time I was doing this.
> 
> What I did here was use the real time window resize event and introduced the logic that is similar to the full screen map resize but with the actual FP screen math and jitter. If you were to build this version you would see the map and elements stay anchored to all window borders (hence the QoL designation).

Ah my bad yeah it is tagged with QoL, okay gotcha makes sense. I'm going to try to see if I can make it so that when the game is in true fullscreen mode that the Map is extended, otherwise not.

React1-X:
> > Ah my bad yeah it is tagged with QoL, okay gotcha makes sense. I'm going to try to see if I can make it so that when the game is in true fullscreen mode that the Map is extended, otherwise not.
> 
> If I remember correctly there is an issue with the fullscreen condition plus how and where the size gets set. I think it's a matter of ifs and elses being badly stacked. Because the math did make sense when I tested.

I pinged you on discord about the fix, but yeah I was able to figure out the true root cause in this PR: https://github.com/bym-refitted/backyard-monsters-refitted/pull/167

I think we're good to close this PR 🙂

PR #128, opened 2025-02-10
fix(pic): wrong other user pic

senninr:
when in map room, the pic for other player is wrong, it shows our own profile pic, should be using pic_square from database for each user


React1-X:
Good spot! Thanks @senninr 

Looks good to me.

React1-X (approved):

PR #134, opened 2025-03-18
[WIP/POC] Bugfix/champion payload issue

tram98:
This is still a POC. We have some bugs left

# Purpose of this PR

This PR addresses an inconsistency in how the server serializes and deserializes champion data, especially affecting local environments using Docker Compose and macOS systems.

## Issue Overview
* In "officially used" environments (Windows/Linux), the server correctly serializes champion data as a JSON string, which aligns with the client’s expectations.
* However, in Docker Compose and macOS, the server sends the data as a JSON array instead of an encoded JSON string, breaking compatibility with the client.
* The exact cause of this behavior could not be identified, so the issue has been circumvented by changing how the data is processed.

## Changes Implemented
* The server now processes champion data as a JSON array rather than handling it as an encoded JSON string.
* Introduced Zod-managed JSON schema for validation and type safety, improving server-side data handling.
* The client has been updated to expect a JSON array instead of an encoded JSON string for champion payloads.
* The base saving mechanism remains unchanged, still sending a JSON string as required by the `UrlLoaderApi` functionality.

## Database Migration Requirement
* This PR requires migration of the `champion` records in the save table.
* Currently, champion data is stored as a JSON string (e.g., `"[{\"foo\":\"bar\"}]"`).
* It must be converted to a JSON array (e.g., `[{"foo":"bar"}]`).
* This can be achieved by a simple Node.js script that:
  * **creates a backup first** ;-)
  * fetches the champion records, 
  * calls `JSON.parse` on them
  * and writes the result back to the database


React1-X:
We got there in the end with this, finally champion data is not shit anymore @tram98 :)

https://github.com/bym-refitted/backyard-monsters-refitted/pull/240

PR #140, opened 2025-03-30
Draft: add mailbox

senninr:


server/src/controllers/mail/getMessageThread.ts
```diff
@@ -0,0 +1,60 @@
...
+    const user: User = ctx.authUser;
+    const { threadid } = GetMessageSchema.parse(ctx.request.body);
+    if (!threadid) {
+      ctx.body = {
+        error: 1,
+        thread: {}
+      };
+      return;
+    }
+    ctx.status = Status.OK;
```
React1-X, line 32: As good practice, we should always have `ctx.status` and `ctx.body` together, status above body. Example:

```js
ctx.status = Status.OK;
ctx.body = {}
```

server/src/controllers/mail/getMessageThread.ts
```diff
@@ -0,0 +1,60 @@
...
+ *
+ * @param {Context} ctx - The Koa context object, which includes the request body.
+ * @returns {Promise<void>} - A promise that resolves when the controller is complete.
+ * @throws {Error} - Throws an error if the request body is missing required fields or if logging fails.
+ */
+export const getMessageThread: KoaController = async (ctx) => {
+  try {
+    const user: User = ctx.authUser;
+    const { threadid } = GetMessageSchema.parse(ctx.request.body);
+    if (!threadid) {
```
React1-X, line 25: When we don't have something that we expect we should use one of our client safe custom errors that we have in `errors.ts`, instead we should throw an error e.g. `throw permissionErr()` - in your case, you might want to create a new one.

server/src/models/thread.model.ts
```diff
@@ -0,0 +1,68 @@
...
+      $or: [
+        { userid: userId },
+        { targetid: userId}
+      ]
+    }, { orderBy: { threadid: "DESC" } });
+
+    return foundThread;
+  }
+
+  public static async findOrCreateThread(threadId: number, userId: number, targetId: number): Promise<Thread> {
```
React1-X, line 49: Ideally, we should move these functions (`findThread` and `findOrCreateThread`) out of the model class, and into something more appropriate such as the `/services` directory.

I think it would make more sense to put them in there, such as `/services/mailbox` since we are doing business logic here.

server/src/models/thread.model.ts
```diff
@@ -0,0 +1,68 @@
...
+    const foundedThread = await this.findThread(threadId, userId);
+    if (foundedThread) {
+      return foundedThread;
+    }
+    if (threadId !== 0 && !foundedThread) {
+      throw saveFailureErr();
+    }
```
React1-X, line 56: This code could benefit from an early return/guard clause - instead of directly returning foundedThread if we have it.

Like so:
```js
if (!foundedThread && thread !== 0) throw saveFailureErr();
```
After this point you know foundedThread will be defined, so there's no need for us to do `if (foundedThread)`

This brings me to a question, do we need to check if `threadId !== 0` here? If we know we won't have a foundedThread? Could this not be simplified to:

```js
if (!foundedThread) throw saveFailureErr();
```

server/src/models/thread.model.ts
```diff
@@ -0,0 +1,68 @@
+import { Entity, Index, OneToOne, PrimaryKey, Property, Unique } from "@mikro-orm/core";
+
+import { v4 } from "uuid";
+import { ORMContext } from "../server";
+import { Message } from "./message.model";
+import { saveFailureErr } from "../errors/errors";
+
+@Entity({ tableName: "thread" })
+@Unique({ properties: ["threadid"] })
+@Index({ properties: ['userid', 'targetid', 'threadid'], name: 'idx_thread_user_target' })
```
React1-X, line 10: For models, can we apply indexes and unique properties on the column-level rather than the class-level? e.g.

```js
@Unique()
@Property()
threadid!: number;
```

Question: here I noticed you are using a Composite Index over regular indexes, are these 3 columns generally always queried together? e.g. `SELECT userid, targetid, threadid FROM.....`

If they are always queried together could you provide some evidence on this? If so, this is okay. Otherwise we can use standard indexes at the column-level.

server/src/app.routes.ts
```diff
@@ -404,4 +410,63 @@ router.post(
...
+
+/**
+ * Report thread
+ * @name POST /api/:apiVersion/player/reportmessagethread
+ */
+router.post(
+  "/api/:apiVersion/player/reportmessagethread",
+  apiVersion,
+  verifyUserAuth,
+  debugDataLog("report message"),
```
React1-X, line 469: For these `debugDataLog` calls, can we provide slightly more descriptive messages? e.g. "Getting Messages Thread", and so on.

React1-X (changes requested):

server/src/controllers/mail/getMessageThread.ts
senninr: done

server/src/controllers/mail/getMessageThread.ts
senninr: the client only expect error field and not showing any message, so if it's error, I has also added errorLog for server

server/src/models/thread.model.ts
senninr: done, move to services/mail/findOrCreateThread.ts

server/src/models/thread.model.ts
senninr: I have simplified this function by moving the thread creation on top before find the thread part

server/src/models/thread.model.ts
senninr: I forgot to update this index, just change it to two composite, and add threadid with unique() and index() too (in total 3 index) choosing composite over normal index because it's also being used for where clause along with orderby clause example:
```
const threads = await ORMContext.em.find(Thread, {
          $or: [
            { userid: user.userid },
            { targetid: user.userid }
          ]
        }, { orderBy: { threadid: "DESC" } });
```

server/src/app.routes.ts
senninr: done, I just use the path as the log message

React1-X:
Looks good to me, I'm going to merge it into develop now and update a couple of things; well done on this feature @senninr <3

Issue #135, opened 2025-03-24
Town Hall downgraded itself to level 1 (base reset)


React1-X:
This bug seems to trigger most commonly at town hall 3 - 5, with the majority seen at town hall 4.
The game client seems to have a check which determines which buildings and the amount of them you should have per town hall, so if the town hall is reset to level 1, it will also remove all buildings associated with higher town hall levels.

Additional evidence suggests that a `rE` property is given to the town hall object, which may stand for 'recycle', suggesting that the game client somehow thinks the town hall was recycled, and therefore resets it.

![Image](https://github.com/user-attachments/assets/9396c77d-30ed-4728-b338-50fdbbff305d)

sir-kick-me2:
Commenting something I found that may relate to the issue of Town Halls downgrading themselves, and most certainly relates to the opposite issue where Town Halls randomly upgrade themselves.

In the client scripts in BASE.as there is a function called `RebuildTH` that appears to be run every time the main yard/inferno yard is loaded. It appears this function is supposed to serve the purpose of dealing with issues relating to the Town Hall being missing or under leveled.

- If there are buildings in the yard that could only exist with a higher level Town Hall, it will automatically upgrade the Town Hall.
(Note: This can only happen if the base has over 2 million "points", which I assume to be a combination of XP plus the value of the buildings inside the yard.)
	
- If the Town Hall is missing, it will attempt to rebuild it.
(Note: The `RebuildTH` function has one parameter that, if set to true, will force a new town hall to be built. However, I don't see anything that ever does that, and even if it was set to true, it would only do something if the base has LESS than 2 million "points".)

In both cases, another function `CaluclateExpectedTownHallLevel` [sic] is called to determine what level the Town Hall should be.

The function is a bit obfuscated since it isn't immediately clear which variables refer to which buildings, so I compiled a list of what each `_loc#_` variable refers to for reference.
```
_loc1_  ---> expectedTownHallLevel
_loc2_  ---> sniperTowersAmount
_loc3_  ---> cannonTowersAmount
_loc4_  ---> infernoCannonTowersAmount
_loc5_  ---> quakeTowersAmount
_loc6_  ---> magmaTowersAmount
_loc7_  ---> currentBuilding
_loc8_  ---> juicerLevel
_loc9_  ---> catapultLevel
_loc10_ ---> yardPlannerLevel
_loc11_ ---> bunkerLevel
_loc12_ ---> hccLevel
[...]

PR #143, opened 2025-04-04
Bugfix: Townhall Up/Downgrading, Saving Unplaced Buildings, Mushrooms Canceling Worker

sir-kick-me2:
Creating a pull request with the intent to resolve issue #135, along with a few other bugs I discovered that ended up being caused by similar issues. I have written down any relevant details regarding each change and the bugs' causes. Let me know if anything here seems wrong, needs clarification, or if you want anything done differently.

The following changes were made:

**BFOUNDATION.as:**  Added a check to `getBuildingSaveData` that excludes `GLOBAL._newBuilding` from being saved to a base's buildingData. This solves an issue where the Town Hall data was being overwritten, causing the game to replace it (Typically with a lower level one).

> Notes:
> - `GLOBAL._newBuilding` is a BFOUNDATION object that refers to the brand new building that the player is attempting to build from the "build" menu. It is cleared when the new building is placed or cancelled.
> - BFOUNDATION objects do not have an ID initialized upon creation, so its ID is treated as 0 until set manually. New buildings built by the player only have their ID set when they are first placed in the yard.
> - This means the building stored in `GLOBAL._newBuilding` will always have an ID of 0. Saving it to a base's buildingData will overwrite whatever building already had an ID of 0. In most cases, this means overwriting the Town Hall.
> - Even if `GLOBAL._newBuilding` always had a unique ID, it still should not be saved, as that would essentially place the building in the yard for free.

**BASE.as:** Added a check to `CalcBaseValue` that excludes `GLOBAL._newBuilding` from increasing a base's Empire Value

> Notes:
> - Empire Value is used to determine outpost takeover costs, and is also added to the player's XP total. Letting buildings affect those values before they are placed (and before resources are expended) does not seem like intended behavior and can potentially be exploited.

**BFOUNDATION.as:** Added a check to `RecycleC` as to not remove a worker from a non-mushroom building if the object being recycled is a mushroom. This solves an issue where workers would stop working on buildings after picking a mushroom, pausing the upgrade/build timer (Usually affecting Town Hall upgrades).

> Notes:
> - All non-mushroom buildings have an ID based on the total buildings in the yard, while a mushroom's building ID is only based on the amount of mushrooms in the yard.
> - The function `QUEUE.Remove` is responsible for removing a worker from a building and freeing them up.
> - The function is called once in MUSHROOMS.as when a mushroom is picked, using the parameter `"mushrooms" + _id`, which tells the code to remove the worker on the mushroom that has that ID. Afterward, It is called a second time in BFOUNDATION.as when the mushroom is recycled, this time using the parameter `"building" + _id`, which tells the code to remove the worker on the non-mushroom building that has that ID. So, when a mushroom is picked that has the same ID as a building that is building/upgrading, the worker on that building will act as if the build/upgrade was canceled and pause its timer.
> - A minor graphical bug still remains: when a matching id mushroom is picked, the "Building/Upgrading" overlay is briefly removed until the building's next Update function is called.
> - Ideally the way buildings are given IDs would be changed to prevent future collision issues.

**BASE.as** Removed erroneous checks in `CaluclateExpectedTownHallLevel` noted in my comments on issue #135. This solves an issue where the Town Hall can suddenly be upgraded when certain conditions are met.

> Notes:
> - This function, along with `RebuildTH`, is called every time a base is loaded to ensure a Town Hall exists and is not underleveled.
> - In the Overworld, the game detected a Town Hall level of at least 4 if there was a Wild Monster Baiter (`_loc13_`) with a level ">= 0", which is always true.
> - In the Inferno, the game detected an Under Hall level of 6 if there were four quake towers built (`_loc5_`) or a level 4+ magma tower (`_loc32_`) existed. Both can be done with a level 5 Under Hall.
> - Ideally this function and `RebuildTH` should no longer be needed, as there should be no situation where a Town Hall is removed or underleveled.

### DIFF
--- a/client/scripts/BASE.as
+++ b/client/scripts/BASE.as
@@ -6359,7 +6359,7 @@ package
             {
                _loc1_ = 3;
             }
-            if(QUESTS._global.brlvl >= 8 || QUESTS._global.b6lvl >= 7 || QUESTS._global.b15lvl >= 3 || QUESTS._global.b23lvl >= 1 || QUESTS._global.b25lvl >= 1 || _loc21_ >= 3 || _loc23_ >= 2 || _loc11_ >= 2 || QUESTS._global.b19lvl > 0 || _loc3_ >= 5 || _loc2_ >= 5 || _loc13_ > 0 || _loc14_ > 0 || _loc16_ > 0 || _loc15_ > 0 || _loc20_ >= 3 || _loc19_ >= 4 || _loc17_ >= 4 || _loc13_ >= 0)
+            if(QUESTS._global.brlvl >= 8 || QUESTS._global.b6lvl >= 7 || QUESTS._global.b15lvl >= 3 || QUESTS._global.b23lvl >= 1 || QUESTS._global.b25lvl >= 1 || _loc21_ >= 3 || _loc23_ >= 2 || _loc11_ >= 2 || QUESTS._global.b19lvl > 0 || _loc3_ >= 5 || _loc2_ >= 5 || _loc13_ > 0 || _loc14_ > 0 || _loc16_ > 0 || _loc15_ > 0 || _loc20_ >= 3 || _loc19_ >= 4 || _loc17_ >= 4)
             {
                _loc1_ = 4;
             }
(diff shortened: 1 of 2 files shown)


React1-X:
Hey @sir-kick-me2 

LGTM! Great work! Very well structured PR, good explanations and changes.
 
My only bit of advice for the future is, don't be afraid to rename some of the local register variables to more meaningful names that are not `_loc_#` - it helps massively for readability and knowing what's happening. 😃

React1-X (approved):

PR #145, opened 2025-04-07
Refactor findTarget function in Krallen, MonsterBase and ChampionBase

MatiasBais:
### Summary

- Unobfuscated `findTarget` for better readability.
- Removed unused variables and irrelevant code (Krallen had flying/burrowing monster logic).
- Optimized targeting logic:
  - Previously: all buildings were collected into an array, then sorted by distance.
  - Now: only tracks the closest building during iteration — no need to sort.

### Performance

- A/B tested locally — no difference observed in pathing behavior.
- Ran attack scenarios (6 times each) on high level Kozus with:
  - 30 Slimes
  - Fomor
  - Krallen  
  → Current version consistently freezes.  
  → This version had zero freezes in all test cases.
- Some lag still occurs (notably using 10m Putty Rage), but the game continues without freezing.

### DIFF
--- a/client/scripts/com/monsters/monsters/MonsterBase.as
+++ b/client/scripts/com/monsters/monsters/MonsterBase.as
@@ -1119,21 +1119,65 @@ package com.monsters.monsters
          this._atTarget = false;
          this._targetCreep = null;
       }
+
+      /*
+       * This function was created to assist in findTargeting
+       * 
+       * @autor: matiasbais
+       * 
+       * @reason: There was a lot of repeated logic in findTarget acrros MonsterBase, ChampionBase and Krallen
+       *          This atemps to make the code a little bit cleaner
+       * 
+       * @param {Object} buildingList - the buildings from which we need the closest ones.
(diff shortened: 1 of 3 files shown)


React1-X:
Hey @MatiasBais 

After some testing with these new changes, it seems as though Krallen's new pathing seems a little off.

I tested with two clients open side-by-side, with and without the changes, attacking Legionnaire, Kozu, Abunakki and Dreadnaught.

Kozu seems relatively fine with some variations, but, when attacking the other tribes, Krallen has a noticeable difference in his pathing when choosing the next building.

- Blue = original pathing
- Red = new pathing changes
- Black = deployment zone of Krallen with initial direction

**Example 1:**

image

After destroying the town hall in this scenario, the new code changes seem to make Krallen more prone to wall-wrapping, rather than targeting the nearest building.

**Example 2:**

image

In this test case, original Krallen goes directly to the Town Hall after destroying that goo factory, while new Krallen decides to wall-wrap and head towards the silo everytime

PR #146, opened 2025-04-08
Auto Bank Resources Counting Towards Player Level

MatiasBais:
### Summary

Removes XP gain from autobanking.

### Details

- This is a **one-line fix** addressing the issue listed in the issues tab.
- Seems like the behavior might have been intentional — it's oddly specific.

Are we certain this wasn’t how it worked in the original game?


React1-X:
Hey @MatiasBais 

It seems this is intentional since it's in the client; the calculation is likely done on the server, or something else is happening, but it seems autobank did in fact give you points. This PR won't be merged, as this is a change that alters the original game client.

PR #151, opened 2025-04-12
Fix Visuals Errors On  HATCHERYCCPOPUP.as

MatiasBais:
## 🐛 Issues Addressed

- Stats bar width was incorrectly displayed for Balthazars.
- Inferno monsters incorrectly showed their cost in Goo.

---

## 🔧 Changes Made

- Unobfuscated all variables and removed unnecessary ones in `MonsterInfoB` for better readability.
- Changed the `speed` variable from `Integer` to `Number` (max speed can be `4.5 km/h`, using `Integer` caused the value to truncate to `4` which led to the display issue).
- Added conditional logic to display `Magma` or `Goo` depending on the monster's type.

---

## 💡 Fun Fact

If you don't have an `HCC`, this error wasn't present because `HATCHERYPOPUP` already uses `Number` for the `speed` variable.

## Issues Remaining

- For inferno monsters it says Goo Cost XXX Magma Cost since that part is render in a different function.

### DIFF
--- a/client/scripts/HATCHERYCCPOPUP.as
+++ b/client/scripts/HATCHERYCCPOPUP.as
@@ -222,16 +222,17 @@ package
       _loc12_ -> damageShown, idk why damage text is render in a different way than the other stats
       _loc2_ -> removed since it does nothing
       _loc4_ -> creatures
-      _loc3_ -> removed since it was not needed, changed to param1
+      _loc3_ -> removed since it was not needed, changed to param1 (creatureID)
       _loc13_ -> level
-      
+      param1 -> creatureID
+
       added v2 var so inferno monsters show their cost in magma not goo
       */
-      public function MonsterInfoB(param1:String) : void
+      public function MonsterInfoB(creatureID:String) : void
       {
          var currentCreature:String = null;
          var damageShown:int = 0;
-         var creature:Object = CREATURELOCKER._creatures[param1];
+         var creature:Object = CREATURELOCKER._creatures[creatureID];
          var speed:Number = 0;
          var health:int = 0;
          var damage:int = 0;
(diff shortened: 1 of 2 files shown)


React1-X:
Hey @MatiasBais 

Could you provide me with a before and after screenshot of these changes? e.g. before the changes vs after the changes; just so I can understand the exact UI fix.

Thanks.

client/scripts/HATCHERYCCPOPUP.as
```diff
@@ -210,111 +210,125 @@ package
...
+      _loc11_ -> currentCreature
+      _loc12_ -> damageShown, idk why damage text is render in a different way than the other stats
+      _loc2_ -> removed since it does nothing
+      _loc4_ -> creatures
+      _loc3_ -> removed since it was not needed, changed to param1 (creatureID)
+      _loc13_ -> level
+      param1 -> creatureID
+
+      added v2 var so inferno monsters show their cost in magma not goo
+      */
```
React1-X, line 230: Let's change the ASDoc comments here to something more like:
```js
/* The following changes have been made to this function
 *
 * @author: matiasbias
 *
 * @description: Renamed local registers to readable variable names and removed unused logic. 
 * Added `v2` variable to conditionally display magma/goo depending on monster type. 
 * Updated speed stat from `int` to `Number` type to handle floating point values.
*/
public function MonsterInfoB(creatureID:String) : void {
```

If you check `MapRoomPopup.as` the function `GetCellsInRange` was rewritten, and we added similar comments for tracking.

React1-X (changes requested):

React1-X (approved):

React1-X (approved):
Cool, looks good to me! Thanks @MatiasBais

PR #153, opened 2025-04-14
Fix sabnox not doubleing damage to favorite building targets

AMXBus:
Since Sabnox has a special type of ranged attack and doesn't function like all the rest of the melee monsters, the attack procedure happens a bit differently. 

As it goes over the FIREBALLS.as class and spawns an entity as a projectile that does the attack. This entity has no check for types of buildings just if it's a building or a monster. 
In order to keep the logic checks to a minimum I have placed the check in Sabnox.as. Since this is the only ground monster that has projectile attacks that do more damage to specific buildings, it seems fitting that it only checks for building type before spawning the projectile per Sabnox attack instead of each time a projectile is called.

*A note - People and the wiki seem to mention that it used to attack towers faster but I was unsure of this so I left the speed as is.

### DIFF
--- a/client/scripts/com/monsters/monsters/creeps/inferno/Sabnox.as
+++ b/client/scripts/com/monsters/monsters/creeps/inferno/Sabnox.as
@@ -7,8 +7,6 @@ package com.monsters.monsters.creeps.inferno
    
    public class Sabnox extends CreepBase
    {
-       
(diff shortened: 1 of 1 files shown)


React1-X (approved):
LGMT! Nice work @AMXBus

PR #154, opened 2025-04-18
Optimizing BRESOURCES, Reducing Long Load Times Caused by Harvesters

sir-kick-me2:
This pull request is an extension of #148. In the PR I had mentioned I had a solution to optimize `BRESOURCES.as` in order to decrease base load times further. This PR includes that solution. As always, let me know if you notice anything wrong, or need any clarification.

### Overview
Harvesters are able to cause long load times for Main Yards (specifically Main Yards). This is due to two reasons:
- There is a while loop in `BRESOURCES.as` that runs when you are loading a Main Yard. Typically, this loop will run once for every 10 seconds since the yard's last save. A level 10 harvester takes ~138,400 seconds to fill from 0, so this loop can theoretically run a maximum of ~13,840 times for each level 10 harvester in the yard, assuming the player had been gone for that long.
- The tickLimit function for harvesters returns a low value tick limit meant for when they are producing, even for Damaged/Destroyed harvesters that can no longer function. Recall that the ProcessC function in `BASE.as` will process the number of seconds from the last save time to the current time, and how many times it runs is based off the value of the tick limit it gets from buildings. The lower the value, the more times it runs.

> [!NOTE]  
> - When not full, harvesters return a tick limit of: productionValue * productionTimeout (resources added to harvester every production interval * seconds per production interval). 
> 	- While productionTimeout by default is 10s, it is also influenced by current health left.
> - A destroyed level 6 harvester, which has a productionValue of 22 and a productionTimeout of 50s, will return a tick limit of 1,100 seconds when its tickLimit function is called.
> - Assuming a base's last save was 30 days ago (2,592,000 seconds), the ProcessC function in BASE.as will run a minimum of 2,356 times when loading a base with a destroyed level 6 harvester (Assuming not being repaired).
>	- This gets even worse with lower level harvesters and harvesters that are just barely under 50% health, as they'd have a lower productionValue and productionTimeout respectively.
> - You can actually see an example of the long load times caused by this in game. There is a user named SinCosPlay in world 2 that has a partially damaged base that has been inactive for a long time. Loading this base took me multiple minutes.

### Changes
**BRESOURCE.as:**
- Added an optional parameter to the produce function to allow it to simulate multiple production iterations at a time without the need of a loop. This is done by simply multiplying the building's "production value" by this parameter.
> [!WARNING]  
> I am aware that SecNum was added as a method to combat cheating, but I am not 100% sure when and how they should be used, which is the one reason why I am unsure of this change in the first place. The `BRESOURCE` class makes use of SecNum, but the optional parameter I am adding isn't a SecNum, and since it essentially bypasses harvester wait times when > 1, I am wondering if this is an issue.

- Modified the Tick function to be able to produce multiple iterations of resources when the building isn't repairing.
- Modified the tickLimit function:
	- Added a check to see if the harvester is currently able to produce.
	- Added a case where If the harvester is not repairing, it will return the total amount of seconds it will take to fill the harvester so all the resources can be produced in a single tick function call.
> [!NOTE]  
> - The case is only for when the harvester is not repairing since productionTimeout is based on the harvester's health, which would become inconsistent and harder to account for when repairing.

### Load Time Comparison
I have attached some videos showing the load times with and without the code changes. 
In these first two videos I am testing load times on an unharmed base. In both videos the following steps are taken:
1. Collect all from the harvesters to ensure they are producing from zero.
2. Leave base.
3. Manually set the base's last `savetime` to 30 days ago in Beekeeper Studio.
3. Reopen base.
The base has a total of 24 harvesters. One of each at levels 1-9, and the rest level 10.

> Flashing images warning for all videos. Clients compiled from VScode are unstable as per the wiki.

Without Changes (Current Patch):
https://github.com/user-attachments/assets/d28892ed-0fcb-4fe0-b8e6-5a8c89d2f854

With Code Changes Applied:
https://github.com/user-attachments/assets/fbfa8b38-e933-4790-810c-5589c567e158

In these seconds two videos I am testing load times on a base with a damaged level 1 harvester at about 15% health left. The base being loaded is the same in both videos, and in both cases I manually set the base's last `savetime` to 30 days ago in Beekeeper Studio.

Without Changes (Current Patch):
I had to cut this video off after a full minute (roughly at 50%).
https://github.com/user-attachments/assets/a698e476-4b9b-437f-a319-2b3d4b5c6017

With Code Changes Applied:
https://github.com/user-attachments/assets/acfc7aad-5791-4b9e-bcbb-2ba310081948


React1-X:
@sir-kick-me2 - very impressive PR, thanks a lot for this change. I went over it and tested it on my end, works great and performance is a lot better! Thanks for the effort on those videos too, it goes a long way! :)

PR #155, opened 2025-04-19
Fink, Bandito, and Eyera Restorations. Wormzors Bugfix

sir-kick-me2:
This pull request does five things, all of which are restoring or fixing bugs related to monster abilities.
- Restores Bandito's Whirlwind ability.
- Restores Fink's Claws ability.
- Restores Eye-ra's ability to deal damage from bunkers without airburst, and partially restores their ability to explode on death.
- Fixes a bug where Wormzor's splash damage was healing enemy monsters.
- Makes king wormzor's splash damage no longer friendly fire buildings.

As always, let me know if anything needs clarification or if you notice any mistakes.

# Overview
Monster objects are able to have components attached to them that allow the monster to gain certain abilities or perform certain actions. For example, On creation of a Fang monster, they get a `PoisonOnAttack` component attached to them if their ability is researched.
- Fink uses the `AOEDamageOnAttack` component for their claws ability.
- Bandito uses the aptly named `BanditoAOEDamageSpin` component for their ability, which extends the `AOEDamageOnAttack` component.
- Both wormzors use the `AOEDamageOnAttackOncePerTarget` component for their splash damage abilities.

All of these components are extensions of the `AOEDamage` component. I have added some more functionality to this component along with adding a few other parameters to a couple Targeting functions and to MonsterBase.

> [!NOTE]
> - The only monster that is a notable exception from this usage of components for abilities is Eye-ra. All of Eye-ra's logic is instead inside CreepBase.as, likely due to the fact they were one of the first monsters added so their logic never got updated.

# Changes
## New Parameters in AOEDamage.as:
**AOEDamage.as:** 
	- Renamed variables/parameters, and added imports for IAttackable and MonsterBase.
	- Added one optional parameter `initialTarget` (default null) to the component's dealAOEDamage function.
	- Added two optional parameters `radiusInner` and `includeInitialTarget` to the component's constructor:
- radiusInner (default 0): Detemines how far away a monster must be before distance-based damage falloff starts to occur.
- includeInitialTarget (default true): Determines if a monster's current target will also take the area damage.

The following all extend the AOEDamage component. They were changed in some way by choice or necessity as a result of this addition:
	- **AOEDamageOnAttack.as**
	- **BanditoAOEDamageSpin.as**
	- **AOEDamageOnAttackOncePerTarget.as**
	- **AOEHealOnDeath.as**
For the most part, I also renamed parameters/variables and added the new optional parameters `radiusInner` and `includeInitialTarget` to their constructors. Some override the dealAOEDamage function, so the optional `initialTarget` parameter had to be added.
> [!NOTE]  
> - These new parameters are so the fink/bandito abilities can deal the same damage regardless of distance, and so they won't deal damage to the same monster/building twice with one attack.
> - AOEHealOnDeath is a component for a rebalaced version of Zafreeti. It is currently unused.

## New Parameters in Targeting.as:
**Targeting.as** Added two optional parameters to function getTargetsInRange to allow it to ignore a specified monster and a specified building. Similarly added a parameter to getBuildingsInRange to allow it to ignore a specified building.
> [!NOTE]  
> - The same kind of parameter already exists in Targeting's getCreepsInRange function, allowing it to ignore a specified monster.
> - This is to add functionality to the new includeInitialTarget parameter in AOEDamage.

## getAllTargets function:
**AOEDamage.as:** Added a new function called getAllTargets that returns an array of every monster and building that should be damaged.
	- Returns a list from Targeting.getCreepsInRange or Targeting.getTargetsInRange
	- Set so friendly monsters (monsters defending from a bunker) won't include buildings as targets so they don't friendly fire their yard. (Only exception being if the friendly monster is specifically targetting a building).
	- Set to not include `initialTarget` parameter in the list if the component's `includeInitialTarget` flag is false.

## _lockRotation Flag:
**MonsterBase.as:** Added a `_lockRotation` flag that turns on/off the ability for monsters to automatically adjust their rotation.
**BanditoAOEDamageSpin.as:** Now sets the new `_lockRotation` flag to true when at their targeted monster, and false otherwise.
> [!NOTE]  
> - Monsters use their `_targetRotation` variable to change their rotation, but any changes made to it would simply get overwritten since the monster's Render function automatically sets `_targetRotation` to be based on where the monster is going.

## Fix Monsters' Targeting Flags:
**Fink.as And Bandito.as:** Fixed/added targeting flags respectively. Also set their AOEDamage component parameters `innerRadius` to 60 (same as outer radius), and `includeInitialTarget` to false.
**CreepBase.as:** Fixed targeting flags for Eye-ra's explosion when defending a base.
> [!NOTE]  
> - Targeting flags are used in Targeting.as to determine whether an attack should hit buildings, attacking monsters, defending monsters, invisible creatures, ect.
> - Targeting flags are added via bitwise operations. The code was using a bitwise AND operation when adding flags, effectively removing all of the flags making abilities target nothing instead. These were changed to bitwise OR operations.

## Eye-ra Exploding on Death:
**CreepBase.as:** 
	- Removed a return statement that was being called before Eye-ra's explosion logic when defending would run.
	- Added a check for attacking creatures at 0 health to call the explode function if they have the exploding tag.
> [!NOTE]  
> - Eye-ra's explosion logic is entirely within CreepBase.as and not in a component. The logic is also entirely different depending on whether eye-ra is attacking a base or defending it. This causes a whole bunch of inconsistencies and bugs, but fixing those is beyond the scope of this pr.

## Other Fixes:
**AOEDamage.as:** When the dealAOEDamage function calls `Targeting.DealLinearAEDamage`, it now uses an absolute value for the damage parameter, since it expects a positive number.
> [!NOTE]  
> - The function used to damage/modify the health of a building _subtracts_ the damage parameter from the current health, while the same function for monsters _adds_ it, which is why `Targeting.DealLinearAEDamage` expects a positive number.
> - This fixes a bug where wormzor's splash damage was actually healing monsters instead of damaging them.

**AOEDamageOnAttack.as:** Inside the onAttack function, Changed a `>` comparison to a `>=` one since it gave abilities a 1 second minimum cooldown even though Bandito and Fink's ability is set to have 0 cooldown.

# In Action
The following are videos demonstrating what the restored abilities look like ingame, as well as before-and-after videos for the wormzors.

> Flashing images warning for some of the videos.

## Fink's Claws:
3 Finks with level 1 Claws against a few octo-oozes:
https://github.com/user-attachments/assets/7091d07d-9282-4c73-8166-377d0fe2b79c

Many more finks with level 3 claws against many itchis with the help of Fomor:
https://github.com/user-attachments/assets/d680611d-94b0-4cdf-96f9-0f5174311ae0

Fink walks up to some walls and smacks multiple of them:
https://github.com/user-attachments/assets/75e61961-986f-49f7-b272-b30984c4eee8
(I thought that was going to be it, but then the fink started hitting both silos right when I stopped recording.)

## Bandito's Whirlwind:
1 Bandito with level 1 Whirlwind against a few octo-oozes:
https://github.com/user-attachments/assets/92f70ff5-56db-4d4a-b566-adc6b612c77e

Many more Banditos against Crabatrons:
https://github.com/user-attachments/assets/5fd3849e-3565-474f-9a8b-baad199b297e

> [!NOTE]  
> - While this ability does not do AOE damage to buildings, it still partially works on buildings since this ability reduces attack cooldown. This effect can be seen in the current patch as well.

## Eye-ra Fixes:
A fearsome enemy is even more fearsome now:
https://github.com/user-attachments/assets/1ad134b4-d549-4b94-b8bb-b71659ac9891

Explosion damage upon death:
https://github.com/user-attachments/assets/0443e26a-8de6-4940-9726-493aee0478bb

## Wormzor Bugfix & Friendly Fire Fix:
Without changes (current patch):
Wormzor: https://github.com/user-attachments/assets/509fa311-5770-4cb1-9b29-35867bfdca20
King Wormzor: https://github.com/user-attachments/assets/44a0fe1c-1bd2-4217-81eb-cac852e3cc69
(Yep, those are green numbers you can see in there.)

With changes applied:
Wormzor: https://github.com/user-attachments/assets/8228b5c5-a814-42b1-9d68-e5722bf54e95
King Wormzor: https://github.com/user-attachments/assets/3065fbc9-15d7-465e-9bd3-73cb1015758f


React1-X (approved):
LGTM! Thanks once again for the thorough explanation and logic changes. :)

PR #160, opened 2025-05-06
Monster Transfer Bugfixes: Bases Swap Monsters, Base Not Showing Monsters

sir-kick-me2:
This pull request is intended to resolve the following issues related to transferring monsters between bases:
1. Often after a transfer, the monsters would then swap between the two bases.
2. After reloading the map after a transfer, the base would appear to have no monsters when selected until the base was loaded by the player again.
3. Transfers do not function when the source cell is outside of the map's rendered cell grid.

As always, do let me know if you spot anything that looks incorrect, need any clarification, or would like something done differently.

# Changes:
### Monsters Swapping After Transfer Fix:
**(Server-Side) transferMonsters.ts:** Server now orders by ascending or descending `baseid` such that the "source" base is always the first result.

**Context:**
- Currently, when the server queries the database for the bases, by default it returns the two bases in ascending order of `baseid`
	- So if the transfer's "target" base had a lower `baseid` than the "source" base, it would treat the "target" base as the "source" base, and the "source" base as the "target" base.

### No Monsters Showing on Base's UI Fix:
**(Client-Side) MapRoom.as:** Outposts now preserve housing and hatchery related data that was previously lost after a monster transfer.

**Context:**
- Data that was lost after a transfer incudes: 
>	- "hcount"		(total hatchery buildings in the base)
>	- "space"			(total housing space in the base)
>	- "finishtime"		(time until housing was finished upgrading)
>	- "overdrivepower"	(multiplier for "hatchery overdrive" powerup)
>	- "overdrivetime"	(time remaining for "hachery overdrive" powerup)
- **MapRoomCell.as** uses these values in order to show which monsters are present in a base.

---

# Changes (May 26th Commits):
### Verifying User Owns Both Bases
**(Server-Side) transferMonsters.ts:** Server now checks if the player owns the source & target base first before performing the transfer.

**Context:**
Being only able to transfer monsters between bases a player owns appears to be the intended behavior of monster transfers, so it would make sense for the server to verify that. This would also prevent an edge case where a player tries to transfer monsters to/from one of their bases that had *just* been taken over by a different player (thus wasn't yet registered on their map).

---
> The rest of the following are changes made to the **MapRoom.as** client-side file.
### Using Copy of Transfer's Source Cell
`_monsterSource` denotes the source cell of a monster transfer; it is the cell where monsters are being transferred from. `_monsterSource` used to be a reference to the visible MapRoomCell object within the rendered 16x14 grid of MapRoomCell objects (18x15 when in fullscreen). When the player moved the map around, inevitably the reference would no longer be pointing to the intended source base, but rather whatever base was stored in the MapRoomCell which *used* to be the intended base.

- Instead of a reference, the `_monsterSource` variable now stores a *copy* of a MapRoomCell object, which won't change as the player moves around the map. Copy is created in the `TransferMonstersA` function.
- The MapRoomCell reference is still stored, however, in a new variable called `_monsterSourceRef`. This is used in the case of a close range transfer where the source cell is still rendered.
- Added a check to `TransferMonstersA` where it will only update the monster data of the source MapRoomCell *reference* if its coordinates are the same as the source MapRoomCell *copy*.

### New Helper Functions
**GetCellZone:** Takes x and y coordinates of a map cell as input and returns an object containing two things: a point object containing coordinates for where the cell's zone begins, and the id of that zone.

**GetPendingZoneRequestIndex:** Takes x and y coordiantes as input and returns the index of the coordinate's corresponding zone in `_pendingMapCellDataRequests`. Returns -1 if not found. In other words, it gets whether or not there is a pending `getarea` server request for the given coordinates, and where it is in the request queue.

**ZoneHasPendingTransferRequest:** Takes a zone id as input and returns true if the `MapRoomCell` objects stored in `_monsterSource` or `_monsterTargetRef` have a zone id that matches the input. Always returns false if `_pendingTransferRequest` is false. In other words, its gets whether or not a map zone has a cell with a pending `transferassets` server request.

**Context:**
- Zones are 10x10 sections of the map, starting from 0x0. The code uses the coordinates of the cell in the zone's top left corner to calculate the id of the zone (id = cell.x * 10000 + cell.y). For example, the cell with coordinates 43x109 would be in the zone 40x100 and that zone's id would be 400100.
- For the sake of consistency the existing `GetCell` function in MapRoom.as was refactored to use the new `GetCellZone` function instead of calculating the zone's point and id on its own. This change doesn't modify its behavior.
- `_pendingTransferRequest` is a new boolean variable. It is intended to be true when the client has an unfinished `transferassets` server request (aka a monster transfer), and false otherwise.
- `_monsterTargetRef` is a new variable storing a MapRoomCell object. It is the target cell of a monster transfer, where monsters are being transferred to. It used to only be stored locally under `targetCell` in the `TransferMonstersA` function. This scope increase is so it can be easily accessed in the new `ZoneHasPendingTransferRequest` helper function.

> [!NOTE]  
> The `_monsterTargetRef` variable *does store a reference* to a MapRoomCell object, which does theoretically mean it can face the same issue as before where it can change when the map is moved by the player. However, buy the time this variable is set, the player is unable to move the map and cannot do so until the monster transfer is finished.

### Modifying Monster Data in `_zones` After Transfer
- After a successful transfer, the transfer's target and source cell monster data is now modified within the `_zones` object. This is so the cell's housed monsters will remain updated, even when reloaded on the map.

**Context:**
The client stores data about all of the map cells it has loaded in an object called `_zones`. This allows the client to easily restore the data of a cell it has already loaded. Currently, this does not get updated after a transfer, so if a player were to reload the cell on the map (without entering that cell's yard), the monster data won't be updated, and the player can transfer the monsters again, effectively duplicating them.

### Waiting For Relevant Server Requests to Finish
- Added logic to the `TransferMonstersA`. If there is a pending `getarea` request that contains the transfer's source or target cells, the function will wait for the request to resolve first before sending the transfer server request.
- Added similar logic to the `RequestData` function. If it would send a `getarea` request for a zone that contains a cell currently undergoing a monster transfer, it will wait for the transfer to resolve first.
- Added a `_priorityMapCellsToRequest` array to act as a "Priority Queue" for loading zones in the map room. When the `RequestData` function sees a map cell in this priority queue, it get that cell's zone and adds/moves it to the front of the `_pendingMapCellDataRequests` queue.
- `TransferMonstersA` now adds the source and/or target cell to the `_priorityMapCellsToRequest` Priority Queue if it needs to wait for `getarea` requests for those cells to resolve. This is so the function will not have to stall as long if the server requests for those cells happened to be far behind in the `_pendingMapCellDataRequests` queue.

**Context:**
The `_zones` object is also modified after a successful `getarea` server request, updating and adding information about the cells it recieves from the server. Since I am now making it so `_zones` is also modifed after a `transferassets` server request, this can lead to a race condition if a cell undergoing a transfer has a pending `getarea` request.

> Example: 
> 1. A `transferassets` request is sent to the server. Map cell 43x109 is the source cell for the monster transfer.
> 2. A `getarea` request is then sent for the zone 40x100, which contains the map cell 43x109.
> 3. The `transferassets` request resolves first, and the monster data for cell 43x109 is updated with the correct number of monsters it should have.
> 4. The `getarea` request is resolved afterward, and the monster data for cell 43x109 is updated with the monsters it had *before* the transfer.

This waiting logic should prevent this race condition from occuring.

> [!WARNING]  
> This waiting logic currently does not have a "time out" period and can theoretically stall forever if a request fails to resolve one way or another.


React1-X:
I seem to still be hitting issues when it comes to transferring monsters from main yards, to outposts which are further away such as on the edge of the main yard's range, or further, with the following error:

Error:
`status":500,"data":{},"internalInfo":"Error: One or both bases not found. From: 2524114639, To: 2524105631`

image

sir-kick-me2:
> I seem to still be hitting issues when it comes to transferring monsters from main yards, to outposts which are further away such as on the edge of the main yard's range, or further, with the following error:
> 
> Error: `status":500,"data":{},"internalInfo":"Error: One or both bases not found. From: 2524114639, To: 2524105631`

After looking into this deeper, this particular issue appears to be a much deeper issue related to how the map room itself is set up.

So, according to MapRoomPopup,as within the `GenerateCells` function, there is only a total of 16x14 MapRoomCell objects ever present/stored on the map at a time (18x15 when in fullscreen).
```
if(GLOBAL._ROOT.stage.displayState == StageDisplayState.FULL_SCREEN)
{
    this._cellCountX = 18;
    this._cellCountY = 15;
}
else
{
    this._cellCountX = 16;
    this._cellCountY = 14;
}
```
These MapRoomCell objects have data about the cell attached to them. Importantly, it includes that particular cell's monster data.
MapRoomCell objects are physically moved to new locations as you move around the map, and the data attached to them gets replaced afterward with data it gets from the server.

There does not appear to be any functionality in the client to _preserve_ any particular MapRoomCell object or the data attached to it.
So in the case of this issue, the game is using the tile that _used_ to be the "from" base (the main yard in your case), but because you moved along the map such that the main yard was outside of the 16x14 / 18x15 range, it got replaced with another tile, in this case, it got replaced with a wild monster cell that had not yet been saved in the database.

I do not know the full extent of the consequences of this quirk with the map room, but could inevitably be the cause of several other issues.

---

As a note, assuming this issue with the map room wasn't present, the issue of monsters getting swapped and not appearing on the UI would still persist since they occur even for closeby transfers where this maproom issue isn't in play. The code changes here _do_ still fix that particular issue with transfers.

React1-X:
Well that's not great... I can see why they might have done this at the time, but seems they never thought about long-traversals across the map...

PR #173, opened 2025-08-05
Fixes to minor Inferno bugs

AMXBus:
These are not functionally related to anything really and can be bundled with any larger update.

- Infernal academy now shows proper text related to monster upgrade requirements and limits (when monsters are max level/when academy needs upgrading and so on)
- D.A.V.E. Club now not part of the UI as there was never any interaction with Inferno and is currently just an exception thrower
- Roughly centered Infernal academy sprite so it doesn't look awkward when around other buildings.

Any other bugs that I may have missed I would gladly addon to this PR as they get reported.

### DIFF
--- a/client/scripts/INFERNOYARDPROPS.as
+++ b/client/scripts/INFERNOYARDPROPS.as
@@ -2364,22 +2364,22 @@ package
             1:{
                //"anim":["anim1.1.png",new Rectangle(11,-2,22,17),44],
                //"anim2":["anim2.1.png",new Rectangle(-32,-49,53,84),44],
-               "top":["top.1.png",new Point(-50,-27)],
+               "top":["top.1.png",new Point(-50,-55)],
                // "shadow":["shadow.1.jpg",new Point(-50,23)],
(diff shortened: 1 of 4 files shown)


client/scripts/INFERNOYARDPROPS.as
```diff
@@ -2364,22 +2364,22 @@ package
             1:{
                //"anim":["anim1.1.png",new Rectangle(11,-2,22,17),44],
                //"anim2":["anim2.1.png",new Rectangle(-32,-49,53,84),44],
-               "top":["top.1.png",new Point(-50,-27)],
+               "top":["top.1.png",new Point(-50,-55)],
```
React1-X, line 2367: Hey @AMXBus I think we can revert these changes for the Infernal Academy - the better solution would be for us to scale the placeholder asset in Photoshop to be not misaligned (e.g. reducing the image size) rather than updating the dimensions in the code, which we may end up coming back to to revert if we find the original asset.

React1-X (approved):

PR #223, opened 2026-02-16
Enable strict mode in AS3 compiler

tram98:
This PR enables the strict mode for the AS3 compiler and resolves all errors that were initially there.

Most errors were related to implicit string -> number coersion caused by the typeless nature of AS3 Dictionaries.

!! IMPORTANT !!
Merge this PR **AFTER** #221, as this branch is based on said PR and cannot work without the changes from #221.

### DIFF
--- a/client/scripts/com/auth/AuthForm.as
+++ b/client/scripts/com/auth/AuthForm.as
@@ -215,7 +215,7 @@ package com.auth
             var formWidth:Number = 450;
             var formHeight:Number = 600;
 
-            languages = KEYS.supportedLanguagesJson as Array;
+            languages = KEYS.supportedLanguagesJson;
             var selectInput:Sprite = createSelectInput();
             contentContainer.addChild(selectInput);
             selectInput.x = 20;


client/scripts/com/auth/AuthForm.as
```diff
@@ -215,7 +215,7 @@ package com.auth
             var formWidth:Number = 450;
             var formHeight:Number = 600;
 
-            languages = KEYS.supportedLanguagesJson;
+            languages = KEYS.supportedLanguagesJson as Array;
```
React1-X: I think we're good to revert this, instead I set the type to an Array in `KEYS.as`, and casted it where we parse the data from the server

client/scripts/com/monsters/monsters/creeps/CreepBase.as
```diff
@@ -1269,7 +1269,7 @@ package com.monsters.monsters.creeps
                   }
                   if(!MapRoomManager.instance.isInMapRoom3 || !BASE.isMainYardOrInfernoMainYard)
                   {
-                     --_homeBunker._monsters[_creatureID];
+                     _homeBunker._monsters[_creatureID] = int(_homeBunker._monsters[_creatureID]) - 1;
```
React1-X: Hey, so just a quick question on this since this has been done in many, many places:

Is the compiler actually complaining about this? I ran the compiler in strict mode and it doesn't seem to be throwing any issues for this? I seem to be only getting 5 actual errors from the compiler.

How are you seeing this?

client/scripts/com/monsters/monsters/creeps/CreepBase.as
```diff
@@ -1279,12 +1279,12 @@ package com.monsters.monsters.creeps
...
                      GLOBAL.player.monsterListByID(_creatureID).unlinkCreepFromData(this);
                   }
-                  --_homeBunker._monstersDispatched[_creatureID];
+                  _homeBunker._monstersDispatched[_creatureID] = int(_homeBunker._monstersDispatched[_creatureID]) - 1;
                   if(_homeBunker._monstersDispatched[_creatureID] < 0)
                   {
                      _homeBunker._monstersDispatched[_creatureID] = 0;
                   }
-                  --_homeBunker._monstersDispatchedTotal;
+                  _homeBunker._monstersDispatchedTotal = int(_homeBunker._monstersDispatchedTotal) - 1;
```
React1-X: Does the compiler complain about this one or did you change it for consistency?

This is declared as `_monstersDispatchedTotal:int = 0` in Bunker.as, I could be wrong but I think this is always initialized and can never be undefined?

I just want to make sure that every cast we do in this PR isn't redundant in some way if we know the value can never possibly be undefined - I think most of the changes you made are fine for the casts, but there might be a couple like this one, and the one I mentioned in `BASE.as` which don't need the cast. (let me know if the compiler complains still)

tram98 (commented):
removed cast to Array

client/scripts/com/auth/AuthForm.as
tram98: done :)

PR #226, opened 2026-02-19
refactor: removed renderer_friend namespace

tram98:
Note: this replaces PR #224, the PR branch was moved from my fork to the main repo.

# Summary

This PR deletes the `renderer_friend` namespace used by the `RasterData` and `Renderer` classes and fixes a critical bug inside the static `RasterData::clear` method, which surfaced following this refactor.

# Context

The `renderer_friend` namespace essentially allowed the two classes to internally share several properties while hiding them to the outside scope.

# Motivation

The reason for this change is somewhat related to my plan to attempt to port the client codebase to Haxe. As that that language does not support custom namespaces, it makes sense to limit the client codebase to a language subset that is shared between AS3 and Haxe.

# Impact

This change should not have any impact on the client behavior. The `renderer_friend` namespace was shared by the two classes mentioned above. As they both share the same package and no other class uses that namespace, migrating the namespace usage to the (package)-`internal` namespace effectively makes no difference.

The change would only be noticable if we started adding classes to the `com.monsters.renderer`- package, as the new classes would then be able to access the internal fields by `Renderer` and `RasterData`.

# Bugfix: `RasterData.clear()` only cleared half of its entries (opening the worldmap seg-faults the flash player on MacOS)

Opening the worldmap calls `BASE.Cleanup()` → `RasterData.clear()`, which could hard-crash the Flash Player on MacOS (it is untested whether other platforms were affected).

The static function `clear()` iterated the registry with `for each` while the instance-function `clear()` splices `this` (itself) out of that same vector:

```as3
for each(_loc2_ in s_rasterData)
{
   _loc2_.clear(param1);   // splices itself out of s_rasterData
}
```

`for each` over a `Vector` uses an incrementing index, so every splice shifts the remaining elements left *past* that cursor. Given `[A,B,C,D,E,F]` it visits `A`, `C`, `E` and stops, so `⌊N/2⌋` entries are never cleared. The trailing `length = 0` then dropped them from the registry regardless.

Those survivors are what crashes: they keep `_cleared == false` and a live `_data`, and `_cleared` is exactly the flag every defensive guard keys on (`Renderer.rasterize`, `set visible`). So a survivor looks valid to every guard while the bitmaps it points at are disposed around it, and it can re-add itself to `s_visibleData` through `set visible` — after which the renderer blits from a disposed `BitmapData`.

The fix drains the vector instead of iterating it, so every entry is visited exactly once:

```as3
while(s_rasterData.length > 0)
{
   rasterData = s_rasterData.pop();
   rasterData.clear(dispose, false);
}
```

`hideDebug()` was given the same treatment. The new `attemptSplice` parameter on the instance `clear()` lets the drain skip the now-pointless `indexOf` lookup, since `pop()` has already removed the entry.

----

# TL:DR

The PR replaces the `renderer_friend` namespace with the `internal` namespace.
This has no effect on the client logic, but makes later migration to Haxe easier.

It also fixes a modify-while-iterate bug in the static RasterData.clear() function that left half of all RasterData instances un-cleared, causing the Flash Player crash when opening the worldmap.

### DIFF
--- a/client/scripts/com/monsters/rendering/RasterData.as
+++ b/client/scripts/com/monsters/rendering/RasterData.as
@@ -7,83 +7,81 @@ package com.monsters.rendering
    import flash.filters.BitmapFilter;
    import flash.geom.Point;
    import flash.geom.Rectangle;
-   
-   use namespace renderer_friend;
-   
+      
    public class RasterData
    {
       
-      renderer_friend static const s_rasterData:Vector.<RasterData> = new Vector.<RasterData>();
+      internal static const s_rasterData:Vector.<RasterData> = new Vector.<RasterData>();
       
-      renderer_friend static const s_visibleData:Vector.<RasterData> = new Vector.<RasterData>();
+      internal static const s_visibleData:Vector.<RasterData> = new Vector.<RasterData>();
       
-      renderer_friend static const s_unsortedData:Vector.<RasterData> = new Vector.<RasterData>();
+      internal static const s_unsortedData:Vector.<RasterData> = new Vector.<RasterData>();
       
-      renderer_friend static const s_debugData:Vector.<RasterData> = new Vector.<RasterData>();
+      internal static const s_debugData:Vector.<RasterData> = new Vector.<RasterData>();
       
-      renderer_friend static var s_needsSort:Boolean;
+      internal static var s_needsSort:Boolean;
       
       private static var s_id:uint;
        
       
-      renderer_friend const _id:uint = s_id++;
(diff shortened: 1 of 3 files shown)


tram98:
@React1-X  I recommend that you go through the PR commit by commit. The first commit removes the renderer_friend namespace, the second and third commits fix the crashes I encountered during my tests and increase the client stability overall.

React1-X (approved):

tram98:
@React1-X I updated the description to include a detailed bug description for future reference.

PR #228, opened 2026-02-19
Haxe: Migrate and enhance SecNum class with testing support

tram98:
Migrate the SecNum class from AS3 to Haxe, improve its implementation, and introduce mock classes for testing. Ensure trace functionality activates only in debug mode.

Depends on #271 and must be merged afterwards.


React1-X:
Closing: decision not to continue with Haxe

PR #247, opened 2026-05-03
fix: subtract looted resources from defender on attack

Chibbluffy:
Attacker loot was being added to the attacker's resources without deducting from the defender, effectively duplicating resources.

Applies to both standard base attacks (baseSave) and inferno attacks (infernoSave). Defender resources are clamped to 0 to guard against the client reporting more loot than the defender actually held.


React1-X:
I think this might need a rebase from main.

```bash
git pull origin main
```

Chibbluffy:
It definitely does lol. I'll get to it tonight or tomorrow

PR #264, opened 2026-08-17
Add empirepoints and savedate to the MR2 bulk snapshot

austin-72:
_Scope reduced per review, with one substitution explained below._

## What

`players[uid]` gains two fields:

```jsonc
"players": { "1234": { "name": "...", "avatar": "...|null", "empirepoints": 614110, "savedate": 1757894400 } }
```

- `empirepoints` — the owner's main-save `points + basevalue`, via the existing `calculateEmpirePoints()`. Level is its position in the experience table (what `calculateBaseLevel()` does); alliance empire points is its sum over members.
- `savedate` — the main save's `savetime` truncated to the UTC day, so it changes at most once per player per day rather than on every save. (ETags still roll on every rebuild while `generatedAt` is in the body — separate question.)

Both come off the existing owners query via `leftJoin("u.save", "save")`. No second query. Cell tuple unchanged; `workerFinish` and the `buildingdata` read are gone.

## Why

Level and last-active day are the two things a map consumer can't get from the snapshot without per-base `/base/load` calls. Neither changes fast, and both are already on the row the owners query reads.

## Changes since last review

- Dropped `workerFinish` and the `buildingdata` column.
- Fields come off the owners query with a join rather than a separate `Save` query.
- `generatedAt` uses `getCurrentDateTime()`; hoisted `now` removed.
- Removed the explanatory comments; what remains is the header contract.
- **`empirepoints` instead of `level`.** Same cost (one int per player), but `level` is a lookup on `empirepoints` that discards the number, and that number is also what alliance rank is computed from. Shipping the input lets a later alliance endpoint carry no stats at all. It changes on the same saves `empirevalue` already does, so no new invalidation. Happy to add `level` alongside if you'd rather have both.

## Files

- `server/src/services/maproom/v2/worldSnapshot.ts`
- `server/src/controllers/maproom/v2/bulk/getSnapshot.ts` — contract comment only

## How tested

Local Docker server (oven/bun 1.4.2, Postgres), MR2 world seeded with 2,500 players (`db:seed:mr2`) plus real accounts.

- `bun run typecheck`: clean.
- Cold rebuild of the 2,500-player world (after a server restart; query + JSON + brotli/gzip): 61 ms on `main`; 67 and 49 ms with the join. Run-to-run variance exceeds any difference between the two.
- `GET /worldmapv2/snapshot` on that world: `players` has 2,500 entries; every entry has numeric `empirepoints` and `savedate`; every `savedate` is a multiple of 86400; every cell tuple still has 11 fields.
- Cross-checked five players against `bym.save` directly (points + basevalue, and savetime truncated to the UTC day): all match, including a real account with points 102,401,264 + basevalue 46,237,308 = 148,638,572 and savetime 1789500483 → savedate 1789430400.
- Level derived from the payload (`empirepoints` against `experiencePoints`) equals the `l` getarea returns for the same player's main yard cell.
- Second request with `If-None-Match` returns 304.


server/src/services/maproom/v2/worldSnapshot.ts
```diff
@@ -106,6 +163,8 @@ const buildSnapshot = async (worldid: string): Promise<WorldSnapshot> => {
       "s.damage",
       "s.protected",
       "s.destroyed",
+      "s.savetime",
+      "s.buildingdata",
```
React1-X, line 167: I think introducing `buildingdata` here everywhere in this endpoint is adding a fair bit of overhead and kinda defeats the purpose of these fast bulk endpoints because adding this column to a query returning hundreds of thousands of rows pulls 1000+ MB of json out of Postgres per rebuild, per world, per minute since buildingdata is such a big json blob per user, and then bun has to parse all 100k+ blobs synchronously on top of that.

I understand you probably want / need the buildingdata so that you can load bases, but don't think this is the endpoint for it

server/src/services/maproom/v2/worldSnapshot.ts
```diff
@@ -143,10 +225,11 @@ const buildSnapshot = async (worldid: string): Promise<WorldSnapshot> => {
       row.damage,
       row.protected,
       row.destroyed,
+      workerFinishTime(row.buildingdata, row.savetime),
     ]);
   }
 
-  const generatedAt = Math.floor(Date.now() / 1000);
+  const generatedAt = now;
```
React1-X, line 232: I think we can remove the hoisted variable you added above, in-line where we compute it and use the util function we already have in the codebase:
```ts
const generatedAt = getCurrentDateTime();
```

server/src/services/maproom/v2/worldSnapshot.ts
```diff
@@ -123,13 +182,36 @@ const buildSnapshot = async (worldid: string): Promise<WorldSnapshot> => {
...
+  // what userCell serves in-game), not a property of whichever outpost row
+  // happens to reference them.
+  const ownerSaves = await postgres.em
+    .createQueryBuilder(Save, "ms")
+    .select(["ms.userid", "ms.points", "ms.basevalue", "ms.savetime"])
+    .where({ userid: { $in: ownerIds }, type: BaseType.MAIN })
+    .execute<OwnerSaveRow[]>("all");
+
+  const mains = new Map<number, OwnerSaveRow>();
+  for (const save of ownerSaves) mains.set(save.userid, save);
```
React1-X, line 196: I dont think we need another query here to postgres just to get those 4 columns, when we can already access them from the owners query just above that I already had, just as a simple join, much better for performance too:
```ts
 const owners = await postgres.em
       .createQueryBuilder(User, "u")
       .leftJoin("u.save", "save")
       .select(["u.userid", "u.username", "u.pic_square", "save.points", "save.basevalue", "save.savetime"])
       .where({ userid: { $in: ownerIds } })
       .execute<OwnerRow[]>("all");
```

server/src/services/maproom/v2/worldSnapshot.ts
```diff
@@ -73,12 +95,47 @@ interface OwnerRow {
...
+  if (!buildingdata) return 0;
+
+  let finish = 0;
+  for (const building of Object.values(buildingdata)) {
+    const countdown = Math.max(Number(building.cB) || 0, Number(building.cU) || 0);
+    if (countdown > 0) finish = Math.max(finish, savetime + countdown);
+  }
+
+  return finish;
+};
```
React1-X, line 137: Same thing here, this whole function only exists to consume the `buildingdata` column, so it comes out with it. 

I think this is an issue of overhead vs the actual gameplay value you get from this, what I mean by value is, realistically you would get maybe 500 out of 130,000+ bases on a world that actually have a builder working right now. The rest are all leftover timestamps from builds that finished long ago since most people build a kit on an outpost and never come back to it, which say nothing a viewer can use, but we still end up shipping all 130,000 of them anyway and make every map refresh like 40% heavier give or take. I think this is something we keep out of these bulk endpoints for now

server/src/services/maproom/v2/worldSnapshot.ts
```diff
@@ -37,6 +41,15 @@ interface CachedSnapshot {
...
   name: string;
   avatar: string | null;
+  /** Map-room level, from the same calculateBaseLevel the in-game map uses. */
+  level: number;
+  /**
+   * Last save time truncated to the UTC day (unix seconds). Day granularity
+   * on purpose: it changes at most once per player per day, so an otherwise
+   * unchanged world keeps an unchanged payload (and ETag) instead of
+   * invalidating on every save.
+   */
```
React1-X, line 51: Can you remove all these AI generated comments everywhere please

server/src/services/maproom/v2/worldSnapshot.ts
```diff
@@ -123,13 +182,36 @@ const buildSnapshot = async (worldid: string): Promise<WorldSnapshot> => {
...
+    const main = mains.get(owner.userid);
+    players[owner.userid] = {
+      name: owner.username,
+      avatar: owner.pic_square,
+      level: main ? calculateBaseLevel(main.points, main.basevalue) : 0,
+      savedate: main
+        ? Math.floor(main.savetime / SECONDS_PER_DAY) * SECONDS_PER_DAY
+        : 0,
+    };
   }
```
React1-X, line 211: Would need to change this too, if using the existing query above, can clean it up to something like:
```ts
for (const owner of owners) {
  const { userid, username: name, pic_square: avatar, points, basevalue, savetime } = owner;

  const level = points && basevalue ? calculateBaseLevel(points, basevalue) : 0;
  const savedate = savetime ? toUtcDay(savetime) : 0;

  const player: SnapshotPlayer = { name, avatar, level, savedate };

  players[userid] = player;
}

const toUtcDay = (timestamp: number): number => Math.floor(timestamp / SECONDS_PER_DAY) * SECONDS_PER_DAY;
```

React1-X:
Thanks for the PR - overall good one! I think its important to distinguish early that these endpoints exist so one computation serves everyone, so anything that scales with total bases rather than active ones works against the whole point of them.

My ask is to split this. `level` + `savedate` I'm happy to merge once they come off the owners query that's already there, that ends up being fewer queries than we run today.

For the whole buildingdata and workerFinish, I'd like left out for now. Detail is in my inline comment, but the short version is that the cost lands on every consumer on every poll, and what it buys is worker state on maybe 500 bases out of 130,000+. I think workerFinish is presented as restoring something that "worked before under getarea", but getarea has never returned buildingdata or savetime,  see `CELL_SAVE_FIELDS` in getArea.ts

Also just a quick note so that we are on the same page: Our [contributing](https://github.com/bym-refitted/backyard-monsters-refitted/blob/main/CONTRIBUTING.md) guide asks that you understand what you're writing and the problem it solves, and that PRs describe how they were tested. Use AI if it helps you, but what you open has to be code you've read, run, and can defend line by line.  At the moment I'm spending longer verifying the claims in the description than reviewing the change itself, and that's the wrong way round.

Otherwise good work!

Thanks @austin-72!

server/src/services/maproom/v2/worldSnapshot.ts
austin-72: Removed buildingdata.

server/src/services/maproom/v2/worldSnapshot.ts
austin-72: Done

server/src/services/maproom/v2/worldSnapshot.ts
austin-72: Done, joined on the owners query

server/src/services/maproom/v2/worldSnapshot.ts
austin-72: Removed

server/src/services/maproom/v2/worldSnapshot.ts
austin-72: Done

server/src/services/maproom/v2/worldSnapshot.ts
austin-72: Done

austin-72:
Alliance snapshot is #283, player lookup is #284. Kept separate as requested.

PR #267, opened 2026-08-20
Fix bun db:init deadlock on fresh database (allOrNothing: false)

### LINKED ISSUE #266: bun db:init deadlocks on fresh DB: CREATE INDEX CONCURRENTLY vs. batched migration transaction
## Summary

Running `bun db:init` against a fresh, empty PostgreSQL database hangs indefinitely partway through `migration:up`, specifically on `20260815_AddUniqueMainSaveIndex`.

## Root cause

`20260815_AddUniqueMainSaveIndex.ts` declares `isTransactional(): false` (required, since `CREATE INDEX CONCURRENTLY` cannot run inside a transaction) and runs:

```sql
CREATE UNIQUE INDEX CONCURRENTLY IF NOT EXISTS save_one_main_per_user
ON bym.save (userid)
WHERE type = 'main'
```

MikroORM's migrator defaults to `allOrNothing: true`, meaning every transactional migration ahead of it in the same `migration:up` run stays open in **one uncommitted transaction** for the duration of the whole run. When the migrator then opens a **separate connection** to run the non-transactional `CREATE INDEX CONCURRENTLY` migration, Postgres has to wait for all transactions that were already in progress when it started — including that still-open batch transaction — before it can proceed. But the batch transaction won't commit until the entire `migration:up` call (including this migration) finishes.

This is a genuine deadlock: neither side can proceed, and the process hangs forever (verified via `pg_stat_activity` — one connection sits `idle in transaction`, the other sits `active`/`Lock`/`relation` on the `CREATE INDEX CONCURRENTLY` statement).

## Repro

1. Fresh clone, fresh empty Postgres DB
2. `bun install`
3. `bun db:init`
4. Process hangs indefinitely once it reaches `20260815_AddUniqueMainSaveIndex` (previous migrations log `Applied` fine)

## Workaround used

Manually killed the stuck `idle in transaction` backend via `pg_terminate_backend`, which let the concurrent index build finish immediately (confirming the deadlock diagnosis). Then had to manually reconcile `bym.mikro_orm_migrations`, since killing that connection rolled back the bookkeeping inserts for every migration in the batch (though the actual DDL had already been applied by `schema:create` for a fresh DB, so the schema itself was fine).

## Suggested fix

Split `db:init`'s `migration:up` step so the non-transactional migration doesn't share a batch with transactional ones ahead of it — e.g. run `migration:up --to <last transactional migration>` first, then a separate `migration:up` call for the rest. Alternatively, set `allOrNothing: false` in the migrator config so each migration commits independently.

Happy to submit a PR for this if a maintainer confirms the preferred approach.

KlaesSoerensen:
## Summary
- Fixes #266 - `bun db:init` hangs forever on a fresh database when it reaches `20260815_AddUniqueMainSaveIndex`
- Root cause: that migration runs `CREATE INDEX CONCURRENTLY`, which cannot execute inside a transaction, so MikroORM runs it non-transactionally on a separate connection. With the migrator's default `allOrNothing: true`, every transactional migration ahead of it stays open in one shared batch transaction until the entire run finishes - but Postgres requires `CREATE INDEX CONCURRENTLY` to wait for all in-flight transactions to finish first. Neither side can proceed.
- Fix: set `allOrNothing: false` in `mikro-orm.config.ts` so each migration commits as it completes, instead of the whole run sharing one transaction.

## Test plan
- [x] Reproduced the hang on a clean database with the original config - confirmed via `pg_stat_activity` that one connection sat `idle in transaction` (holding the batch open) while another was `active`/blocked on a relation lock running `CREATE INDEX CONCURRENTLY`
- [x] Applied the fix, reset to a fresh database again, ran `bun db:init` - completed cleanly in seconds, all 20 migrations recorded in `mikro_orm_migrations`, and `save_one_main_per_user` confirmed present via `pg_indexes`
- [x] Repeated the before/after comparison a second time to confirm it wasn't a fluke
- [x] Restarted the dev server against the resulting database and confirmed it serves requests normally

### DIFF
--- a/server/src/mikro-orm.config.ts
+++ b/server/src/mikro-orm.config.ts
@@ -55,5 +55,11 @@ export default defineConfig({
     path: path.join(import.meta.dirname, "./database/migrations"),
     pathTs: path.join(import.meta.dirname, "./database/migrations"),
     glob: "*.ts",
+    // Must be false: a non-transactional migration (CREATE INDEX CONCURRENTLY in
+    // 20260815_AddUniqueMainSaveIndex) cannot proceed while an earlier transactional
+    // migration in the same batch is still open on another connection - Postgres
+    // makes CONCURRENTLY wait for every in-flight transaction to finish, and the
+    // batch transaction only commits after the whole run succeeds. That's a deadlock.
+    allOrNothing: false,
   },
 });


React1-X:
Hi @KlaesSoerensen 

Thanks for the PR - this fixes the hang, but I don't think it's the right fix. It changes global migration semantics to accommodate one migration, meaning with this fix, now on a mid-run failure in the `db-init` container leaves the DB half-migrated. That's a downgrade, for one file.

I'll push a fix for the issue in a few minutes, thanks for looking into it! (:

PR #268, opened 2026-08-20
[experimental] Show building health in More/Upgrade popups

KlaesSoerensen:
## Summary
- Adds a **Health** line to the building "More" popup, showing the building's current max HP for its level
- Adds a **Health** line to the "Upgrade" popup, showing the current -> next-level HP and percent increase (e.g. `Health: Increases from 500 to 950 (90%)`) - this format already existed for walls specifically; this generalizes it to every building type
- Health is treated as a numeric attribute, grouped with other per-level stats (Production/Capacity, Range/Damage) exactly the way those already group with each other, rather than always being its own separate block
- Adds `client/scripts/com/monsters/utils/HtmlTextUtils.as` with a small `EnsureBreaks(html, count)` helper that normalizes section spacing (strip trailing `<br>`, then guarantee exactly `count`) - used at every section boundary (title/description -> numeric attributes -> requirements) instead of relying on each building's own upgrade-description text to happen to leave the right number of trailing breaks, which previously produced inconsistent spacing (missing or doubled breaks) depending on building type
- Town Hall is a deliberate special case: its upgrade description is prose about unlocks, not a numeric stat delta, so Health gets its own section there (double break) instead of grouping with it

Feature tag: `experimental`

## Test plan
- [x] Compiled the debug client and verified in-game against a local server across multiple building types: resource buildings (Twig Snapper, Goo Factory), towers (Tesla Tower, Black Diamond Spurtz Cannon), walls, Housing, and Town Hall
- [x] Confirmed via Flash trace instrumentation (temporarily added, removed before this PR) that the exact HTML produced matches the intended section-spacing rule for each case, not just visual eyeballing
- [x] Confirmed walls' existing HP-upgrade display is untouched (still uses their own code path) and doesn't get a duplicate Health line
- [x] Confirmed Requirements section still renders correctly (with or without prerequisite bullets) alongside the new Health line


React1-X:
Hi @KlaesSoerensen 

Really cool feature thanks! However, this project is a faithful restoration of the original game, with the aim being a 1:1 re-implementation. Since this feature was not in the original game, I will be closing this PR - but we can archive it, if anything comes up in the future on another branch for a custom server, thanks!

PR #270, opened 2026-08-23
Fix yard planner logging players out mid-rebuild

KlaesSoerensen:
## Summary
- The AFK/session timeout (`GLOBAL.AFK()`) logs a player out via `POPUPS.Timeout()` after 10 minutes without the session timer being reset.
- That reset normally happens on large stage mouse movement, with a couple of existing spots (`SPECIALEVENT.as`, `SPECIALEVENT_WM1.as`) explicitly calling `GLOBAL.UpdateAFKTimer()` for legitimate engagement that doesn't move the mouse much.
- The yard planner never did this: `BasePlannerPopup.onDesignStateChange` fires on every place/move/delete in the planner but never reset the timer, so players doing careful small-movement layout work for 10+ minutes were logged out and lost their in-progress rebuild.
- Added `GLOBAL.UpdateAFKTimer()` to `onDesignStateChange` so every yard planner modification now keeps the session alive like other actions do.

## Test plan
- [x] Open the yard planner and place/move/delete buildings continuously (small mouse movements only) for 10+ minutes; confirm no AFK timeout/logout occurs.
- [x] Confirm normal AFK timeout still triggers when genuinely idle (no input) for 10 minutes.

### DIFF
--- a/client/scripts/com/monsters/baseplanner/popups/BasePlannerPopup.as
+++ b/client/scripts/com/monsters/baseplanner/popups/BasePlannerPopup.as
@@ -774,6 +774,7 @@ package com.monsters.baseplanner.popups
       public function onDesignStateChange(param1:Event = null) : void
       {
          this.hasBeenSaved = false;
+         GLOBAL.UpdateAFKTimer();
       }
       
       public function Remove() : void


tram98 (approved):
Looks reasonable to me.

This change effectively resets the AFK timer every time a change occurs inside the yard planner. This is only client side and does not trigger any requests to the server, hence no risk there.

@KlaesSoerensen Thank you very much for your contribution :)

@React1-X feel free to merge.

PR #273, opened 2026-08-27
Cumulative resource-only wall upgrade

KlaesSoerensen:
## What

Adds a second **"Upgrade All (Resources)"** button on the wall info panel, next to the existing Shiny "Upgrade All". It opens the STORE popup (same layout as the Shiny version) with one row per town-hall-allowed wall tier — Stone / Metal / Gold / Black Diamond — each showing the **total resource cost** to raise every lower wall up to that tier, and the number of walls affected.

Picking a tier is all-or-nothing: if the player can afford the whole pool the resources are spent immediately and **one worker walks wall-to-wall** applying each upgrade as a single continuous job that is *not* bound by the 5-slot worker-queue cap. In-progress work survives a logoff.

The existing Shiny "Upgrade All" button and the `BLK2`–`BLK5` packages are untouched — all new `STORE.as` code is additive `RBLK*` branches.

## Why

This is a preservation project with no plans to sell Shiny, so bulk wall upgrading as a Shiny sink no longer serves a purpose — it was just tedious clicking (each wall upgrade is only 5 seconds). This replaces the busywork with a resource cost while keeping the Shiny option available for anyone who wants it.

## How it works

| Concern | Approach |
|---|---|
| Cost | `BulkWallUpgrade.buildPlan(tier)` sums each wall's per-level step costs (`_buildingProps.costs`) from its current level up to the tier into one pool |
| Spend | `BASE.Charge` per resource, once, on confirm — ships as the normal save delta |
| Worker | `BulkWallJob` reserves one worker via `WORKERS.Assign` and advances it wall-to-wall; never enters `QUEUE._stack` |
| Persistence | each planned wall gets `_countdownUpgrade` (serialises as existing `cU`) plus a new optional `bwT` key in `buildingdata`; `BulkWallJob.resumeFromLoad()` rebuilds the job on base load. No schema change — `BuildingData` already allows unknown keys |
| Town-hall gating | tiers above the allowed wall level are not shown (same table the Shiny packages use) |

## Files

- **New:** `client/scripts/BulkWallUpgrade.as` (plan/cost/affordability/store rows), `client/scripts/BulkWallJob.as` (the one-worker job + load resume)
- `client/scripts/BUILDINGINFO.as` — the new button
- `client/scripts/STORE.as` — additive `RBLK` render + purchase branches
- `client/scripts/GLOBAL.as` — `BulkWallJob.tick()` in the 1-second loop
- `client/scripts/BASE.as` — resume on load, cancel on cleanup
- `client/scripts/BFOUNDATION.as` — `_bulkWallTarget` field + `bwT` export/import
- `server/src/game-data/store/storeItems.ts` — `RBLK2`–`RBLK5`, `RBLK2I`, `RBLK3I` entries
- `server/src/types/BuildingData.ts` — doc-comment for `bwT`
- `server/public/gamestage/assets/*.json` — new localisation keys (English; other languages carry English placeholders)

## How tested

Manually in the debug Flash build against a local server (no automated test suite exists for the client). Verified: button + tier screen render, per-tier cumulative cost and wall count, all-or-nothing affordability + shortfall message, purchase → confirm → resources charged → one worker sweeping walls up through the tiers, "already running" guard, Shiny path unchanged.

Not yet exercised on this branch: reload mid-job (persistence path is wired but unverified end-to-end), no-idle-worker wait, mode-change pause/resume, outpost/inferno yards. Reviewers with those setups, please poke at them.

## Anti-cheat coordination (needs a maintainer)

On the open-source server `validateSave` is a no-op stub, so this works for local dev. **Production runs the private anti-cheat module.** A tier upgrade makes one `/base/save` carry an unusual shape — many walls gaining `cU` at once plus one large resource drop — which a naive detector could flag. Someone with private-anti-cheat access should confirm a bulk wall upgrade won't trip a ban before this ships live. If it can't be reconciled, the fallback is to make each tier apply instantly (loop `Upgraded()` like the Shiny path) and drop `BulkWallJob` entirely.

## Tags

`experimental`

🤖 Generated with [Claude Code](https://claude.com/claude-code)


React1-X:
Hi thanks for the contribution, really appreciate the effort here.

Unfortunately I can't merge this into `main`. BYMR is a preservation project, and the goal from the start has been a 1:1 restoration that stays as faithful to the original as possible, so balance changes fall outside that scope.

That said, this isn't wasted work. I'm archiving the PR in case we ever put together a rebalanced version of the game, and you're welcome to re-open it against a `rebalanced` branch or in a fork.

We also have the custom-servers channel on Discord for exactly this kind of thing - showing off custom servers and rebalance ideas. Would be great to see it there. :)

One more thing worth raising. It's clear this was generated end-to-end with an AI coding tool, and I want to be upfront that this is a problem for us regardless of scope. Our [contributing guidelines](https://github.com/bym-refitted/backyard-monsters-refitted/blob/main/CONTRIBUTING.md) ask contributors to understand what they're writing and the problem it solves, and to include a description covering what changed, why it was needed, and how it was tested. That's hard to satisfy when the change wasn't authored by hand.

AI tools are fine as an assistant, but the code you submit needs to be code you've read, understood, and tested yourself. We maintain a small codebase with a very specific goal, and reviewing generated changes nobody can explain costs us more time than it saves

PR #274, opened 2026-08-29
Attack battle reports — PR 1a (data pipeline)

KlaesSoerensen:
## Attack battle reports — PR 1a (data pipeline)

The launcher's **Attack Logs** page has a per-row **"View Details"** button that
currently opens a *"not yet available"* modal, because no structured data backs it.
Today the attack is simulated on the attacker's Flash client, which uploads only a
**localized HTML string** (`ATTACK.LogRead()`); the `attack_logs` row is created at
attack start with `loot: {}` / `attackreport: {}` and never backfilled.

This PR adds the **data pipeline** so a real data sheet has something to render.
The launcher UI is a separate follow-up PR against `bymr-desktop-launcher`.

### What it does

- **AS3 client** — a new all-static `AttackReport` accumulator collects a structured
  report *beside the existing `ATTACK.Log()` call sites* during the sim: per-affected-building
  type/pos/HP/loot, attacking force (monsters flung + champion), siege weapons, catapult
  power-ups, overall damage %, duration, outcome. Serialized on attack-end into a compact,
  enum-coded JSON string (`saveData.battlereport`).
- **Server** — `BaseSaveSchema` gains a `battlereport` field; `baseSave` validates it against
  a zod schema with hard anti-bloat caps, correlates it to the `attack_logs` row via a new
  nullable `attackid` column (migration included), and stores the compact blob in the existing
  `attack_logs.attackreport` jsonb column. `attack_logs.loot` (long unused) is populated too.
- **New endpoint** — `GET /api/:apiVersion/attacklogs/:id` returns a decoded, friendly
  `AttackReportView` (enum ids expanded to names), authorized to the attacker or defender only,
  Redis-cached. The existing list endpoint gets a field projection so it no longer ships the
  full report.

### Scope / deliberate omissions (→ a follow-up PR 1b)

`b[].dd` (defensive damage dealt), `b[].k` (kills), `b[].ml` (bunker/champion monsters lost),
`atk.ml` (attacker losses) are **not** produced here — they need combat-path attribution
threaded through `PROJECTILES.Spawn` → `PROJECTILE` and a separate `CreepBase` path, which
roughly doubles the client surface and is the most fragile part. The schema keeps these fields
optional; the decoder fills them `0` / `[]`, so the launcher renders those columns as zero
until 1b lands.

### Design principles

- **Best-effort**: a missing / malformed / oversized / uncorrelatable report never fails the
  player's base save, never throws, never bans. Building damage was already client-authoritative;
  the server validates *shape and size* only.
- **Enum tables** (`attackReportEnums.ts` / `.fixture.json` / AS3 `AttackReportEnums.as`) are
  byte-identical, append-only, index = wire value; a test guards the drift.
- Compact JSON in `jsonb` + Postgres TOAST — no extra compression lib.

### Testing

- `bun test` (new to `server/`): enum-drift guard, schema cap rejection, decoder round-trip —
  14 tests.
- `server/scripts/verify-battle-report.ts`: exercises `persistBattleReport` + `decodeAttackReport`
  against local Postgres/Redis incl. the malformed / mismatched-`attackid` negative cases — 18/18.
- **Manual E2E**: verified against a real MR3 flatten through the Flash client —
  `GET /attacklogs/:id` returns the decoded view with correct flung-monster counts, champion,
  per-building loot, and outcome; list endpoint confirmed to exclude the report body.
  `server/scripts/setup-mr3-attack-test.ts` (dev-only) sets up an attackable opponent.
- Not yet exercised manually: siege/catapult paths, partial-damage (hp > 0), retreat outcome,
  MR2/inferno/tribe (tribe attacks legitimately produce no report → `{ report: null }`).

### Known follow-ups

- Building `x`/`y` are screen-pixel coords, not grid coords (sanctioned v1 shortcut).
- `attack_logs.loot` is the attacker-client's reported total; the server separately caps the
  defender's actual resource transfer, so the report can overstate the loss (pre-existing
  game loot-accounting).
- `?filter=<invalid>` still caches under an unbustable key (low-severity, self-inflicted).

### Relation to the #183 POC

This supersedes the stalled POC in #183 (`feature/attack-logs-reports`). That branch correlated attack-end by *watching `save.attackid`* — first a 1-second polling loop, then Postgres triggers + SQL functions — and @React1-X pushed back (2025-11-17):

> "the complexity has now shifted into the database layer … for the sake of consistency on the server we should try [to] avoid [triggers/functions]. A lightweight, server-side event emitter is still the cleanest, most explicit, and most maintainable approach long-term."

This PR follows that direction:

- **No DB triggers or functions.** The migration is a single `ADD COLUMN attackid` + one index; all processing is plain TypeScript.
- **No polling / no watchers.** The report is persisted synchronously in `baseSave` when the client's end-of-attack `over` save arrives — the attack-ended event *is* that request.
- **Consistent with the codebase.** `persistBattleReport(...)` is called inline from `baseSave.ts` exactly like the existing `attackLootHandler` / `championHandler` / `monsterUpdateHandler`.

**One deliberate deviation from the literal wording:** it is a direct `await` call, not an `EventEmitter` + listener. Reasons — there is no event-emitter/pub-sub pattern anywhere in the server today (an emitter would be the inconsistency React1-X wants to avoid), and `persistBattleReport` must run *inside* `baseSave`'s unit of work (persist the row before the flush, bust the cache after), which needs an awaited inline call, not a fire-and-forget event. Happy to add an emitter layer if the maintainers prefer it.

Also: #183's core motivation — *"`save.attackreport` being overwritten on every attack"* — is solved here for the structured data, which now lives on the per-attack `attack_logs` row and is never overwritten. The HTML `save.attackreport` stays ephemeral, which is fine: it only feeds the in-game post-attack popup, which only needs the current attack.
🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/client/scripts/ATTACK.as
+++ b/client/scripts/ATTACK.as
@@ -628,6 +628,7 @@ package
                   _flungSpace.Add(CREATURES.GetProperty(_loc8_,"bucket") * _flingerBucket[_loc8_].Get());
                   _loc7_ = KEYS.Get(CREATURELOCKER._creatures[_loc8_].name);
                   _loc6_.push([_flingerBucket[_loc8_].Get(),_loc7_]);
+                  AttackReport.RecordFlung(_loc8_,_flingerBucket[_loc8_].Get());
                   _loc13_ = 0;
                   while(_loc13_ < _flingerBucket[_loc8_].Get())
                   {
--- a/client/scripts/ATTACK.as
+++ b/client/scripts/ATTACK.as
@@ -159,6 +159,7 @@ package
          _flingCount = 0;
          _log = [];
          _attackStart = GLOBAL.Timestamp();
+         AttackReport.Reset();
          _flungSpace = new SecNum(0);
          _loot = {
             "r1":new SecNum(0),
@@ -1027,6 +1028,7 @@ package
          BucketClear();
          if(!_sentOver)
          {
+            AttackReport.RecordOutcomeRetreat();
             if(BASE._saveOver != 1)
             {
                BASE.Save(1,false,true);
--- a/client/scripts/AttackReport.as
+++ b/client/scripts/AttackReport.as
@@ -13,6 +13,7 @@ package
    {
       private static var _buildings:Object;
       private static var _lootTotal:Array;      // [r1,r2,r3,r4]
+      private static var _flung:Object;         // creatureID string -> cumulative int count flung this attack
       private static var _siege:Object;
       private static var _catapult:Object;
       private static var _retreated:Boolean;
@@ -23,6 +24,7 @@ package
       {
          _buildings = {};
          _lootTotal = [0, 0, 0, 0];
+         _flung = {};
          _siege = {};
          _catapult = {};
          _retreated = false;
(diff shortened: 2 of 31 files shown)


React1-X:
Closing this one.

This is substantially AI-generated - the structure, the file layout, and the approach don't reflect any understanding of how this part of the codebase works, and several conventions we use throughout are ignored. Per [CONTRIBUTING.md](https://github.com/bym-refitted/backyard-monsters-refitted/blob/main/CONTRIBUTING.md), pull requests a maintainer judges to be sub-par and substantially AI-generated are closed without further review.

We've covered this with you before, so I'll keep it short: AI is fine for repetitive or mechanical work. It is not fine for whole files, features, or architectural decisions. If you can't explain why every line is there, don't open the PR. The PR description clearly states it was generated with Claude Code.

I appreciate the time and effort you're putting in to build these things out, but programming / software engineering is more than just getting something to work with an AI tool and calling it a day. It may be better to provide your ideas in the suggestions channel on our discord, or contribute to minor things.

KlaesSoerensen:
fair, my bad, wont do again.

PR #278, opened 2026-09-10
Refactor: Moved academy and related classes to their own packages

tram98:
This PR is the start of my attempts to increase readability and portability of the AS3 codebase.

## Motivation

The client codebase is a mess, partly caused by the decompilation, but of course also by the high velocity from the beginnings of the game. This makes analysis of the codebase significantly more challenging. This PR attempts to improve the package structure of the codebase.

The objective is to group the classes inside the top-level package into a new appropriate package structure to improve the ease of navigation. This also makes the coupling of different components more transparent.

This PR is the first step towards that objective by moving all classes related to the Academy Building and Logic into appropriate packages under a new `com.bymrefitted` namespace. The purpose of the new namespace is to distinguish between the classes already migrated and the ones still to migrate.

## Additions
- Created a new package `com.bymrefitted.academy` for academy logic
- Created a new package `com.bymrefitted.buildings` for all buildings (those with `extends BFOUNDATION`) and their respective `*hit` classes

## Changes
- Moved ACADEMY, ACADEMYPOPUP and ACADEMYPOPUP_CLIP to `com.bymrefitted.academy`
- Renamed `ACADEMY` to `AcademyService` (new pattern to apply to all future logic-classes)
- Moved `BUILDING26` and `building26hit` to `com.bymrefitted.buildings`
- Renamed `BUILDING26` to `AcademyBuilding` (new pattern: name building classes not by id, but by their actual function)
- Renamed `building26hit` to `AcademyBuildingHit`
- Updated all references to the moved classes and added `import`-statements to all affected classes
- Reduced visibility of many properties to minimum level required

## Lessons learned
- Functions may be dynamically invoked by other classes. Deleting functions just because the language server does not report any usages is DANGEROUS. Example: `frame.as` (the class handling ui window creation) supports some standard actions (such as closing the window) and dynamically looks up handler functions (such as `Hide`) on the `this.parent` property. If the function does not exist, nothing is invoked. Removing a function here would break UI functionality.

### DIFF
--- a/client/scripts/ACADEMYPOPUP.as
+++ b/client/scripts/ACADEMYPOPUP.as
@@ -1,5 +1,6 @@
 package
 {
+   import com.bymrefitted.academy.AcademyService;
    import com.monsters.display.ImageCache;
    import com.monsters.managers.InstanceManager;
    import flash.display.Bitmap;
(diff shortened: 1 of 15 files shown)


tram98:
@copilot I fixed the description. The correct building type is indeed 26, not 23. Anything else?

PR #279, opened 2026-09-12
refactor(client): replaced string-based UI action lookup with type-safe interface casts

tram98:
## Summary

Replaces the client's runtime, string-based UI action lookup with type-safe interface casts, so the contract between a window and its frame is enforced by the compiler instead of by naming convention.

## Background

A *frame* is the window chrome: it draws the window background and border, and owns the actions common to every window — fullscreen, help, and hide/close. There are four variants (`frame`, `frame1`, `frame2`, `frame3`).

Frames are always used as a **child of the popup they decorate**. That is by design, and it is what makes their use of `parent` meaningful: when the user clicks the frame's close button, the frame walks back up to `parent` — the popup that owns it — and asks that popup to close itself. The same goes for help and fullscreen.

The problem was how the frame decided whether its parent could handle an action. It probed for a method by *name*, at runtime:

```as3
if("Hide" in parent)
{
   (parent as MovieClip).Hide();
}
```

Nothing about that is visible to the compiler. Renaming or removing `Hide()` on a popup silently breaks that popup's close button, with no build error and no obvious failure until someone clicks it. In the other direction, any class that happens to expose a member called `Hide` — for whatever unrelated reason — becomes a valid target by accident. The relationship between a frame and the window hosting it was real, load-bearing, and completely undeclared.

## The change

Three interfaces now declare that contract explicitly, under `com.bymrefitted.ui.interfaces`:

| Interface | Method | Implementers |
| --- | --- | --- |
| `IHideActionHandler` | `Hide(param1:MouseEvent = null)` | 72 |
| `IHelpActionHandler` | `Help(param1:MouseEvent = null)` | 8 |
| `IFullScreenActionHandler` | `FullScreen(param1:MouseEvent = null)` | 1 |

Every window the frame chrome previously reached by name now declares which of these actions it supports, and the frames cast instead of probing:

```as3
var handler:IHideActionHandler = parent as IHideActionHandler;
if(handler != null)
{
   handler.Hide(param1);
}
```

The help button's visibility check in `frame2`/`frame3` becomes `parent is IHelpActionHandler` in place of `"Help" in parent`.

Each implementing window gains the optional `MouseEvent` parameter on its handler to satisfy the interface signature. `IMapRoom.Hide()` was widened to `Hide(param1:MouseEvent = null)` for the same reason, keeping it consistent with its implementers.

No `"Hide" in`, `"Help" in` or `"FullScreen" in` lookups remain anywhere in `client/scripts`.

## Behaviour

Behaviour-preserving. Two details worth knowing:

- The frame now forwards the originating `MouseEvent` to the handler, where it previously called `Hide()` with no arguments. The parameter is optional and currently ignored by every implementer, so direct calls such as `QUESTS.Hide()` are unaffected.
- `frame.as` previously cast via `(parent as Object).Hide()`, which succeeded against *any* parent object. It now fails the cast cleanly and falls through to `_customCloseFunction` when the parent is not an `IHideActionHandler` — the behaviour `frame1`/`frame2`/`frame3` already had.

Classes that merely happen to own an instance method named `Hide` are deliberately left untagged — for example `INFERNOPORTAL`, which is a `BFOUNDATION` building rather than a window, and was never an intended close target. Making the contract explicit is precisely what lets us exclude them.

## Testing

- The client compiles.
- Manually tested across a broad range of general windows: close, help and fullscreen buttons all behave as before, and the help button still appears only on windows that actually implement `Help`.

## Note on AI usage

🤖 PR description Generated with [Claude Code](https://claude.com/claude-code), reviewed by me.

I used GitHub Copilot and Claude during the implementation of this PR.

I manually implemented the new interfaces, their usage inside the `frame` classes and updated the first 10 POPUP classes to implement the respective interfaces and optionally patch the method signatures of the affected methods.
GitHub Copilot then used my changes as a reference to implement the rest. I reviewed all changes myself, fixed a few errors and then tested the client UI in depth.

### DIFF
--- a/client/scripts/ACADEMYPOPUP.as
+++ b/client/scripts/ACADEMYPOPUP.as
@@ -1,5 +1,7 @@
 package
 {
+   import com.bymrefitted.ui.interfaces.IHelpActionHandler;
+   import com.bymrefitted.ui.interfaces.IHideActionHandler;
    import com.monsters.display.ImageCache;
    import com.monsters.managers.InstanceManager;
    import flash.display.Bitmap;
@@ -7,7 +9,7 @@ package
    import flash.display.DisplayObject;
    import flash.events.MouseEvent;
    
-   public class ACADEMYPOPUP extends ACADEMYPOPUP_CLIP
+   public class ACADEMYPOPUP extends ACADEMYPOPUP_CLIP implements IHideActionHandler, IHelpActionHandler
    {
       
       public static var _page:int = 1;
--- a/client/scripts/ALLIANCEPOPUP.as
+++ b/client/scripts/ALLIANCEPOPUP.as
@@ -1,5 +1,6 @@
 package
 {
+   import com.bymrefitted.ui.interfaces.IHideActionHandler;
    import com.monsters.alliances.ALLIANCES;
    import com.monsters.alliances.AllianceConstants;
    import com.monsters.alliances.AllianceTabBase;
@@ -12,7 +13,7 @@ package
    import flash.display.MovieClip;
    import flash.events.MouseEvent;
 
-   public class ALLIANCEPOPUP extends MovieClip
+   public class ALLIANCEPOPUP extends MovieClip implements IHideActionHandler
    {
       private static const W:int = 860;
       private static const H:int = 580;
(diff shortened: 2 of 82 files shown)


tram98:
@React1-X This is ready for review

